PR #2, opened 2024-11-28
feat: use base16 colors

GeoffreyFrogeye:
Currently jjui seems to use hardcoded True Color.
This changes to use hardcoded ANSI 16 colors, like jj itself. This allows for respecting the user's color scheme, including light mode support. This doesn't match `jj` default colors, but should keep the existing jjui colors for you.

### DIFF
--- a/internal/ui/common/styles.go
+++ b/internal/ui/common/styles.go
@@ -3,14 +3,14 @@ package common
 import "github.com/charmbracelet/lipgloss"
 
 var (
-	Black      = lipgloss.Color("#000000")
-	Cyan       = lipgloss.Color("#8be9fd")
-	Pink       = lipgloss.Color("#ff79c6")
-	Yellow     = lipgloss.Color("#f1fa8c")
-	Red        = lipgloss.Color("#ff5555")
-	Green      = lipgloss.Color("#50fa7b")
-	Comment    = lipgloss.Color("#6272a4")
-	Foreground = lipgloss.Color("#f8f8f2")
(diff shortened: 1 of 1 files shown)


idursun (approved):
Awesome! Thank you very much!

net:
@idursun for some reason this doesn't seem to work anymore. I'm on jjui 0.8.8 with Ghostty, and colors 0-255 don't use my terminal's configured colors for that range.

net:
Ah, I missed that this PR was for the default colors, not configurable colors. That said, it seems configured colors use exact ANSI256 colors instead of the configured terminal colors.

Here's an example of another tool using the configured terminal colors for 0-255 in lipgloss: https://github.com/sst/opencode/commit/d127a1c4ebe326344dc77fe3d136c033da6031fd#diff-1329f5e85891a0199861978dc66ce2932efc2bb563f5c7b50442a08a159ea565R238

idursun:
@net I am using base16 colours here: https://github.com/idursun/jjui/blob/5234755c4c6462025c12e3c8786d961b9522f91f/internal/ui/common/styles.go#L7-L24

The only colour that's not base16 is the highlight background colour which can be changes through the configuration to be base16, ANSI256, or TrueColor.

I am also using ghostty as my terminal and haven't seen any problems with the themes. Can you share your ghostty configuration? Also, can you make report this as a separate issue so that folks who might be having similar problem can chime in?

net:
@idursun done :) #105

PR #6, opened 2024-12-21
fix: Add missing abandon keymap help to short help row

Adda0:
This PR adds the missing keymap short help for the abandon operation. The remaining operations should already be present in the short help row.


idursun (approved):
Thanks!

PR #14, opened 2025-02-10
rename modules to allow remote go install

nickchomey:
I would like to install JJUI via `go install` (and, actually via [mise's Go backend](https://mise.jdx.dev/dev-tools/backends/go.html)https://[mise.jdx.dev/dev-tools/backends/go.html](https://mise.jdx.dev/dev-tools/backends/go.html)) but I cant because the module is named just `jjui` rather than `github.com/idursun/jjui`

This PR should allow for installing via either.

### DIFF
--- a/README.md
+++ b/README.md
@@ -89,7 +89,12 @@ You can build `jjui` from source.
 ```
 git clone https://github.com/idursun/jjui.git
 cd jjui
-go install cmd/jjui.go
+go install cmd/jjui/main.go
+```
+
+### From go install
+```
+go install github.com/idursun/jjui/cmd/jjui@latest
 ```
 
 ### From pre-built binaries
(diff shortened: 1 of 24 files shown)


idursun (approved):
:shipit: Thank you very much for this!

nickchomey:
Wow, that was quick! Thanks! Ive got it installing with mise now, which will allow me to more easily manage my tools, as well as update jjui when there's new releases

Adda0:
Thank you for the fix as well. This also fixes the installation through Nix, where we had to patch this manually during the build process. Though, since I have only a rudimentary knowledge of the Go ecosystem, I was not aware of this convention.

nickchomey:
Youve done great work for someone who doesn't know about Go!

PR #23, opened 2025-03-06
fix(ui): allow 'p' and '?' when editing a bookmark

kkmlr:
Hey, great tool. However, when creating bookmarks the main Update func is still operating on some keys (such as 'p' and '?'). This fixes it, but it might not be the best way. Feel free to discard if it's not right.

### DIFF
--- a/internal/ui/revisions/revisions.go
+++ b/internal/ui/revisions/revisions.go
@@ -2,6 +2,8 @@ package revisions
 
 import (
 	"bytes"
+	"slices"
+
 	"github.com/charmbracelet/bubbles/help"
 	"github.com/charmbracelet/bubbles/key"
 	tea "github.com/charmbracelet/bubbletea"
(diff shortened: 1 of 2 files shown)


idursun:
Hey, thanks for the report and putting the effort to provide a fix. 

I totally missed that!

Technically, what you have done here would work but I believe there's a better way:

You can just implement the `Focusable` interface on the `SetBookmarkOperation` so that the revisions widget can report that the current operation is in focused mode so that the key press messages are only sent to the revisions and nothing else.

In short, the following change in `set_bookmark_operation.go` should do the trick:

```diff
+func (s SetBookmarkOperation) IsFocused() bool {
+       return true
+}
```
I can wait for you to make the change, or if you don't want to bother then I can do it myself. Up to you!

kkmlr:
ahh, ok, cool. I got a bit lost in all the models to be honest and didn't see it was as simple as that. You can make the change since you know the code much better and what else it might affect. I'll close the PR. 
Again, thanks for a great tool.

PR #24, opened 2025-03-12
feat: use git root as location

kkmlr:
Since almost everyone is using colocated jj, I think it makes sense to use the git root as the default location. 
I use jjui in a tmux popover which uses the current working directory, so I'm only rarely in the git root directory when I run it.

Feel free to close if unwanted or request changes.

### DIFF
--- a/cmd/jjui/main.go
+++ b/cmd/jjui/main.go
@@ -3,7 +3,9 @@ package main
 import (
 	"fmt"
 	"os"
+	"os/exec"
 	"runtime/debug"
+	"strings"
 
 	"github.com/idursun/jjui/internal/config"
 	"github.com/idursun/jjui/internal/ui/context"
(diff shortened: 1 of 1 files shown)


idursun:
Hey, thanks for this! That's a great idea.

I have one question though:

> Since almost everyone is using colocated jj, I think it makes sense to use the git root as the default location.

Well, I don't. However, can we generalise this into something "almost everyone using jj is using `jj` cli". In this case, doesn't `jj root` do the same thing and work even if it is not git colocated? I might be misinterpreting the command though. Let me know if it is not the jj equivalent.

I get the following when I run `jj root`:
```
[I] ~/repositories/jjui/cmd
❯ jj root
/Users/idursun/repositories/jjui
```

kkmlr:
using `jj root` as the command is obviously better. I didn't even know that command existed (I'm new to jj). You want me to make the change? If so, ok to squash?

> Well, I don't

Out of curiosity, how do you sync it with github then?

idursun:
> I didn't even know that command existed (I'm new to jj)

Me too! Also, if it is not already obvious, I am new to golang too.

> You want me to make the change? If so, ok to squash?

Please go ahead with the change, I want to merge your PR as this was your idea in the first place!

> Out of curiosity, how do you sync it with github then?

I use `jj git push`, or from `jjui` I press `g` and then `p`. Having said that though, what `jjui` supports at the moment is not enough for all use cases. One can push just a bookmark with `jj git push --bookmark myexperimental-branch` or all the bookmarks with `jj git push --all`, etc. I am working on a change to account for all these use cases in the new version, which looks like something like the following:

image

kkmlr:
Oh, OK, I thought the git commands only worked with a git colocated repo, but looking at the docs again looks like it's only needed to make git tooling to work, cool.
Change pushed.

idursun (approved):
Nice one, thanks!

PR #50, opened 2025-04-07
Adding nix flake.

vic:
jjui is already on nixpkgs. However as jjui evolves, it might be possible that people might want to try a fresh release without having to wait on it being available on nixpkgs. We now provide a simple flake for just that.

jjui can be run directly with:

```shell
nix run github:idursun/jjui
```

or added as an input on their flakes.

We also provide a minimal development shell for those wanting to contribute back to jjui.


idursun:
This is nice! Thanks!

PR #60, opened 2025-04-22
chore(nix): Add flake check GH action.

vic:
Just to make sure that jjui is buildable with nix.

Previously, if we added a new Go dependency, we
wont notice the vendored-hash mismatch, since CI was not testing with nix.

Now we make sure that the package builds correctly under nix.

### DIFF
--- a/.github/workflows/nix.yml
+++ b/.github/workflows/nix.yml
@@ -1,0 +1,15 @@
+name: nix flake check
+
+on:
+  pull_request:
+  push:
+    branches: [main]
+
+jobs:
+  flake-check:
+    name: nix flake check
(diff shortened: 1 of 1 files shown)


vic:
Here's the build checking that the exposed jjui package is built.

image


btw, this action just checks for x86_64-linux (runner is ubuntu-latest), don't know if we should enable other platforms (at least macos-latest). There's nothing platform specific on jjui, so I guess we are safe just checking on linux (for speed and not consuming more resources), but If you think that we should also check on macos-latest let me know.

idursun:
Nice! Thanks for this. Yes, just checking against ubuntu-latest should be fine for now 👍🏿

PR #74, opened 2025-05-04
feat: add an '(empty)' message to the diff view for empty diffs

### LINKED ISSUE #69: Feature: show some sort of message for empty diffs in the diff view
I just tried to `describe` an empty change, but I mistakenly pressed `d` instead of `D`, unwittingly opening the diff view. 
Since I didn't know what that was, upon being presented with a blank screen, I thought the program had crashed or something.

I think it would be nice to display something like `(empty)` to explain to the user what's happened, maybe additionally something like "esc to go back". 

We could additionally or alternatively show the commit header with the Ids and Author, as in the preview pane.

It seems like the relevant code is here
https://github.com/idursun/jjui/blob/d8a2592a448d70b3ccb0c13826de9e37f163c63e/internal/ui/diff/diff.go#L35-L41

If this idea sounds good, I can have a go at implementing.

sullyj3:
closes #69 

@idursun you mentioned some other stuff I could look into, I'll just submit this for now and possibly or possibly not submit a follow-up depending on whether I can figure it out

### DIFF
--- a/internal/ui/diff/diff.go
+++ b/internal/ui/diff/diff.go
@@ -34,7 +34,11 @@ func (m Model) View() string {
 
 func New(output string, width int, height int) tea.Model {
 	view := viewport.New(width, height)
-	view.SetContent(output)
+	content := output
+	if content == "" {
+		content = "(empty)"
+	}
+	view.SetContent(content)
 	return Model{
 		view: view,
 	}


idursun:
Thanks! That's absolutely fine.

PR #75, opened 2025-05-05
feat: add status bar to diff view

sullyj3:
Todo:
- [ ] set state of status bar correctly
  - [ ] set mode
  - [ ] set help
    - [ ] refactor diff.Model to have a keymap? not totally sure yet


sullyj3:
that failing test in ci, `TestModel_Update_SplitsSelectedFiles`, passes on my machine, and doesn't seem like it should be relevant to what I'm doing?

https://github.com/idursun/jjui/blob/509a61940d978a4c22691f7b1cbfc18b9236185b/internal/ui/operations/details/details_test.go#L48-L63

internal/ui/diff/diff.go
```diff
@@ -1,21 +1,39 @@
...
 
 type Model struct {
-	view viewport.Model
+	view   viewport.Model
+	keymap config.KeyMappings[key.Binding]
+}
+
+func (m *Model) ShortHelp() []key.Binding {
+	// TODO: not sure how to get the full list of mappings that are passed through
+	// to the viewport (eg d scrolls down, u scrolls up)
```
idursun: They should be available through `viewport`s keybindings. We use the defaults here.

idursun:
Thanks for working on this! It's looking really good.

internal/ui/ui.go
```diff
@@ -206,7 +206,14 @@ func (m Model) Update(msg tea.Msg) (tea.Model, tea.Cmd) {
 
 func (m Model) View() string {
 	if m.diff != nil {
-		return m.diff.View()
+		m.status.SetMode("diff")
+		m.status.SetHelp(m.diff.(*diff.Model))
+		footer := m.status.View()
+		if diff, ok := m.diff.(*diff.Model); ok {
```
idursun: You already figured a way but you can also change the type of `diff` in the ui's model to `*diff.Model` to make your life easier like:

```diff
diff --git a/internal/ui/diff/diff.go b/internal/ui/diff/diff.go
index 37c4ee424f..83f39b8000 100644
--- a/internal/ui/diff/diff.go
+++ b/internal/ui/diff/diff.go
@@ -25,7 +25,7 @@
 	return [][]key.Binding{m.ShortHelp()}
 }
 
-func (m Model) Init() tea.Cmd {
+func (m *Model) Init() tea.Cmd {
 	return nil
 }
 
@@ -33,7 +33,7 @@
 	m.view.Height = h
 }
 
-func (m *Model) Update(msg tea.Msg) (tea.Model, tea.Cmd) {
+func (m *Model) Update(msg tea.Msg) (*Model, tea.Cmd) {
 	switch msg := msg.(type) {
 	case tea.KeyMsg:
 		switch msg.String() {
diff --git a/internal/ui/ui.go b/internal/ui/ui.go
index a76eb4e1f9..01458bcffc 100644
--- a/internal/ui/ui.go
+++ b/internal/ui/ui.go
@@ -32,7 +32,7 @@
 	previewModel            *preview.Model
 	previewVisible          bool
 	previewWindowPercentage float64
-	diff                    tea.Model
+	diff                    *diff.Model
 	state                   common.State
 	error                   error
 	status                  *status.Model
@@ -207,12 +207,10 @@
 func (m Model) View() string {
 	if m.diff != nil {
 		m.status.SetMode("diff")
-		m.status.SetHelp(m.diff.(*diff.Model))
+		m.status.SetHelp(m.diff)
 		footer := m.status.View()
-		if diff, ok := m.diff.(*diff.Model); ok {
-			footerHeight := lipgloss.Height(footer)
-			diff.SetHeight(m.height - footerHeight)
-		}
+		footerHeight := lipgloss.Height(footer)
+		m.diff.SetHeight(m.height - footerHeight)
 		return lipgloss.JoinVertical(0, m.diff.View(), footer)
 	}
 ```

File: internal/ui/diff/diff.go (hunk 1 above)
Comment (idursun) [reply internal/ui/diff/diff.go]: something like the below:

```go
km := m.view.KeyMap
return []key.Binding{km.Up, km.Down, km.HalfPageDown, km.HalfPageUp, km.PageDown, km.PageUp, m.keymap.Cancel}
```

internal/ui/ui.go
sullyj3: Yeah, that makes sense. I did that in one of my exploratory commits, but I got into a weird state at some point and abandoned a bunch of commits and forgot to put it back that way.

sullyj3: I'm still in the process of learning about go's memory management. I didn't realise escape analysis was a thing and you could just return pointers to stuff on the stack, so that held me back for a bit.

idursun: Yes, I am also not much of a go expert, and it took a while for me to figure it out. That's why `diff` was left as `tea.Model` because at the time I thought it was the only way that something can be set as nil. 

I am winging it as I go.

internal/ui/diff/diff.go
```diff
@@ -1,25 +1,44 @@
...
+	m.view.Height = h
+}
+
+func (m *Model) Update(msg tea.Msg) (*Model, tea.Cmd) {
 	switch msg := msg.(type) {
 	case tea.KeyMsg:
-		switch msg.String() {
-		case "esc":
+		switch {
+		case key.Matches(msg, m.keymap.Cancel):
```
idursun, line 41: ❤️

idursun (approved):
Thanks! Nice work.

sullyj3:
Cheers, no worries

Issue #72, opened 2025-05-03
Add `jjui` to homebrew


idursun:
This is needed at some point but I have no idea how to go about it. 

Any help is appreciated.

lvignoli:
I have [a working formula](https://github.com/Homebrew/homebrew-core/compare/master...lvignoli:homebrew-core:jjui?expand=1) against [lvignoli/jjui](https://github.com/lvignoli/jjui).
Once a new release incorporates #81, I'll contribute it to homebrew/core and notify you @idursun.

It should then be easy to update, as per https://docs.brew.sh/Formula-Cookbook#updating-formulae.

idursun:
Tyvm for working on this! #81 is merged now.

I can create a release pretty soon.

lvignoli:
Awesome, let me know when you cut a new release and I'll update the branch and open the Homebrew PR 😀

idursun:
@lvignoli I have released [0.8.8](https://github.com/idursun/jjui/releases/tag/v0.8.8) just now

lvignoli:
jjui being added to Homebrew at https://github.com/Homebrew/homebrew-core/pull/223195

lvignoli:
jjui v0.8.8 now available on Homebrew: https://formulae.brew.sh/formula/jjui ✨

idursun:
Thank you very much for this and for #84 🎉

PR #81, opened 2025-05-11
fix: allow build version to be set from ldflags

lvignoli:
When building from a standalone VCS-free tarball of the repo, info.Main.Version is `(devel)`, not `""`.
This fix first check if main.Version has been set by the build options, then default to get the version info embedded by go.

This is necessary for  #72.


idursun:
Awesome! Thanks. I guess this clarifies #80 too.

PR #85, opened 2025-05-16
Update instruction on installation for archlinux-based linux distros

TeddyHuang-00:
This PR updates the README to include installation instruction for archlinux-based distros.

Thank you for the great work of `jjui`. I just created an AUR package for `jjui` so that archlinux-based distro users can easily install it as well: https://aur.archlinux.org/packages/jjui-bin. I will maintain this package as long as needed.

### DIFF
--- a/README.md
+++ b/README.md
@@ -86,6 +86,16 @@ The latest release of `jjui` is available on Homebrew core:
 brew install jjui
 ```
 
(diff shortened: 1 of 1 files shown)


idursun (approved):
Hey @TeddyHuang-00 thank you very much for this!

PR #97, opened 2025-05-29
feat: add quick describe operation

Jollywatt:
This is a feature request in the form of a minimal working example.

The inline text interface for setting bookmarks is unexpectedly cool. I thought it would be nice to have something similar for setting short descriptions, bound to `enter`. Of course, you can still use a text editor for longer messages, but this is also convenient!

### DIFF
--- a/internal/config/keys.go
+++ b/internal/config/keys.go
@@ -18,6 +18,7 @@ var DefaultKeyMappings = KeyMappings[keys]{
 	Quit:             []string{"q"},
 	Undo:             []string{"u"},
 	Describe:         []string{"D"},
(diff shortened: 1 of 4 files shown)


idursun:
Oh brilliant work 💯 

I don't know if you knew but this used to exist in `jjui` in the first releases and I was planning to bring this back. 

jjui_description

I'll have a look and let you know. Thanks again!

Jollywatt:
> I don't know if you knew but this used to exist in jjui

Oh great! Feel free to leave this unmerged then.

I think I prefer the behaviour of your implementation where you type over, not under, the existing description

PR #102, opened 2025-06-03
feat: Add shortcuts for common `git {push/fetch}` operations

### LINKED ISSUE #101: nitpick: New `Git Operations` UI needs more keystrokes
## Overview

I recently updated to the new version of `jjui`, which introduced a UI for git operations. I appreciate that it gives more flexibility as to what is being pushed (i.e. pushing only the hovered bookmark or all bookmarks in the current revset.)

However as a minor nitpick, it requires more keystrokes than the previous version with less optionality. Rather than just `g + p`, now it's `g + p + <j/k/up/down> + enter`.

Image

### Request

While it's nice to have more fine-grained control over what `git push` / `git fetch` operation is being executed, it would be great for each option in the presented list to have a hotkey associated with it, rather than having to navigate with `j/k` or arrow keys and then pressing enter.

Current: `g + <p/f> + <j/k/up/down> + enter`
Requested: `g + <p/f> + <hotkey-for-option>`

clabby:
## Overview

Adds three new shortcuts for the most common git operations:
1. `g + p + r` - Pushes the bookmarks in the current revset
2. `g + p + a` - Pushes all bookmarks
3. `g + f + f` - `git fetch`

Also renames the `Push` / `Fetch` keys to `PushFilter` / `FetchFilter` respectively, since these are no longer actions.

Note that this is a breaking change to the configuration.

closes #101


idursun:
Hey, thanks for working on this!

This is looking good. However, I usually try a few alternative approaches before settling on a change like this. 

Do you mind if I keep this open until I test a couple of ideas I have in mind? I can link you to the branches when I have something to show to get your opinion if you like.

clabby:
> Hey, thanks for working on this!
> 
> This is looking good. However, I usually try a few alternative approaches before settling on a change like this.
> 
> Do you mind if I keep this open until I test a couple of ideas I have in mind? I can link you to the branches when I have something to show to get your opinion if you like.

Yeah of course - no rush! For now I'll use the local build :)

clabby:
Any updates here @idursun ?

internal/config/keys.go
```diff
@@ -69,9 +69,12 @@ var DefaultKeyMappings = KeyMappings[keys]{
...
 	Git: gitModeKeys[keys]{
-		Mode:  []string{"g"},
-		Push:  []string{"p"},
-		Fetch: []string{"f"},
+		Mode:        []string{"g"},
+		PushFilter:  []string{"p"},
+		PushRevset:  []string{"p", "r"},
+		PushAll:     []string{"p", "a"},
+		FetchFilter: []string{"f"},
+		Fetch:       []string{"f", "f"},
```
idursun, line 77: It's not very clear from the way this keys work but they don't define the sequential key presses, but instead define alternative keys to invoke that item. Only one of them will be processed depending the order of handling in the `Update` function.

internal/ui/git/git.go
```diff
@@ -85,10 +85,16 @@ func (m *Model) Update(msg tea.Msg) (tea.Model, tea.Cmd) {
...
+		case key.Matches(msg, m.keymap.Git.PushFilter):
 			return m.filtered("push")
-		case key.Matches(msg, m.keymap.Git.Fetch):
+		case key.Matches(msg, m.keymap.Git.FetchFilter):
 			return m.filtered("fetch")
+		case key.Matches(msg, m.keymap.Git.PushRevset):
+			return m, m.context.RunCommand(jj.GitPush(), common.Refresh, common.Close)
+		case key.Matches(msg, m.keymap.Git.PushAll):
+			return m, m.context.RunCommand(jj.GitPush("--all"), common.Refresh, common.Close)
+		case key.Matches(msg, m.keymap.Git.Fetch):
```
idursun, line 96: Menu items have their command argument attached to them, which is used to render the menu item title. We should be invoking whatever command args are attached to them. Doing it this way is fine, but it will end up duplicating the logic.

idursun (changes requested):
Left comments about the default keybindings, the way they are handled and also the usage of `RunCommand`.

To make the same bindings work, then I think you need to do something like the below:

* if `f` is pressed, and if not in `fetch` filter then it is `fetch filter`
* if `f` is pressed, and if `fetch` filter active, then it is `git fetch`

idursun:
@clabby I left a review here. Sorry for the wait. I played with a couple of ideas similar to what you tried to achieve in this PR but then I got lost while trying to make those nicely fit into the rest of the code base. Because, what we do here should be easily applicable to `bookmarks` and `custom commands` popup windows.

Once this works as intended, then I am happy to merge this and expand it to the rest of the popup windows.

PR #108, opened 2025-06-05
perf: make the preview pane more responsive when switching commit

glehmann:
the debounce of 200ms made it feel a bit laggy. I tried 100ms, and while it was significantly better, it still didn't felt as responsive as expected. 50ms feels just right, and it doesn't seem to overload my system.


idursun (approved):
Thanks! That works for me too!

Issue #128, opened 2025-06-16
Nix flake stopped building on macOS


ilyagr:
I'm getting an error when running `nix build .` on `aarch64-darwin` since commit 95663e4. The previous commit compiles fine. Running `nix flake update --flake .` makes no difference.

Here's the build log:

```
Running phase: unpackPhase
@nix { "action": "setPhase", "phase": "unpackPhase" }
unpacking source archive /nix/store/dv08s3cqyj6idmic9zqn6ab9n5wy43zy-z87q8rgpjz9gks3cdrk42fbvkl0mqwch-source
source root is z87q8rgpjz9gks3cdrk42fbvkl0mqwch-source
Running phase: patchPhase
@nix { "action": "setPhase", "phase": "patchPhase" }
Running phase: updateAutotoolsGnuConfigScriptsPhase
@nix { "action": "setPhase", "phase": "updateAutotoolsGnuConfigScriptsPhase" }
Running phase: configurePhase
@nix { "action": "setPhase", "phase": "configurePhase" }
Running phase: buildPhase
@nix { "action": "setPhase", "phase": "buildPhase" }
Building subPackage ./cmd/jjui
go: inconsistent vendoring in /private/tmp/nix-build-jjui.drv-0/z87q8rgpjz9gks3cdrk42fbvkl0mqwch-source:
	github.com/charmbracelet/bubbletea@v1.3.5: is explicitly required in go.mod, but not marked as explicit in vendor/modules.txt
	github.com/charmbracelet/colorprofile@v0.3.1: is explicitly required in go.mod, but not marked as explicit in vendor/modules.txt
	github.com/charmbracelet/x/exp/golden@v0.0.0-20241011142426-46044092ad91: is explicitly required in go.mod, but not marked as explicit in vendor/modules.txt
	github.com/charmbracelet/bubbles@v0.21.0: is explicitly required in go.mod, but not marked as explicit in vendor/modules.txt
	github.com/charmbracelet/x/ansi@v0.9.2: is explicitly required in go.mod, but not marked as explicit in vendor/modules.txt
	golang.org/x/sync@v0.15.0: is explicitly required in go.mod, but not marked as explicit in vendor/modules.txt
	golang.org/x/sys@v0.33.0: is explicitly required in go.mod, but not marked as explicit in vendor/modules.txt
	golang.org/x/text@v0.26.0: is explicitly required in go.mod, but not marked as explicit in vendor/modules.txt
[...]

PR #132, opened 2025-06-18
CI and Nix: make `nix flake check` build the flake

ilyagr:
Previously, it merely checked that the flake has correct syntax.

See also <https://github.com/idursun/jjui/issues/128>.

--------

This does slow down the CI, but only a little (the run for this PR took 1m8s, hopefully this will be somewhat consistent).

Instead of this PR, another option would be to rename the check to make it clear that it checks the flake syntax only (as it does without this PR).

I also don't know much about Nix, so there might be a better way of implementing this; this implementation was a guess based on `nix flake check` docs.

### DIFF
--- a/nix/default.nix
+++ b/nix/default.nix
@@ -20,6 +20,7 @@
     {
       packages.default = jjui;
       packages.jjui = jjui;
+      checks.default = jjui;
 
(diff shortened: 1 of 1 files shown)


idursun (approved):
Thanks! Let's give it a go.

PR #150, opened 2025-06-30
build(nix): add git version to --version

teto:
before:
```
➜ jjui --version     
(devel)
```

after
```
result/bin/jjui --version
0b5d3eb8e800989ed91845e131e49b8f3bacd21a
```

even better would be if I could retreive the next version to print `jjui 0.9-nightly+REV` or even the current version.

### DIFF
--- a/nix/default.nix
+++ b/nix/default.nix
@@ -7,11 +7,15 @@
   };
 
   perSystem =
-    { pkgs, ... }:
+    { pkgs, lib, ... }:
     let
-      jjui = pkgs.buildGoModule rec {
+      self = inputs.self;
+      version = if (self ? rev) then self.rev else "dirty-${self.dirtyRev}";
+      jjui = pkgs.buildGoModule {
         name = "jjui";
-        src = ./..;
+        src = lib.cleanSource ./..;
+
+        ldflags = [ "-X main.Version=${version}" ];
         vendorHash = builtins.readFile ./vendor-hash;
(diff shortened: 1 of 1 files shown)


Adda0 (approved):
You beat me to it. Looks good to me.

If you really wanted, you could retrieve the current latest released version from GitHub releases, but I think this is already enough. You generally do not need this additional information, only the commit for debugging purposes, bug reports, etc.

idursun (approved):
Thanks!

teto:
> If you really wanted, you could retrieve the current latest released version from GitHub releases,

I would like to do the build in the sandbox without internet access/manual changes.

I would like to add the last release because a git commit does't tell you much. I would like this kind of output for instance and ideally reflect this in the nix store path as well

```
nvim --version
NVIM v0.12.0-nightly+2b4c112
```

I can't find any jjui version number in go.mod , is there some file I could parse ?

Adda0:
There is none since the project sets the version not from `go.mod`, but from build flags during the build process from GitHub releases, as seen in the [build script](https://github.com/idursun/jjui/blob/7a6b379327d7e3d00857e44913d34e45eb927816/.github/workflows/release.yml#L34). You would have to download the version from the GitHub releases into a new source file, and during build time parse this file, AFAIK.

ilyagr:
I made a small fixup to this in https://github.com/idursun/jjui/pull/152, feel free to take a look.

Issue #151, opened 2025-06-30
Empty revsets


whatisaphone:
jjui doesn't handle empty revsets well. Two ways to repro:

1. `jjui -r 'none()'`
	- expected behavior: empty list
	- actual behavior: it shows "loading" forever
2. `jjui`, then hit 'L' and change the revset to `none()`
	- expected behavior: empty list
	- actual behavior: the old list remains on screen

Maybe instead of a blank screen, it could show "(no matching revisions)" or similar

PR #154, opened 2025-07-02
fixed display of empty revsets (#151)

Gogopex:
just a quick fix for #151:

- `jjui -r 'none()'` now returns "(no matching revisions)"
- `jjui`, then hit 'L' and change the revset to none() now displays "(no matching revisions)"

Screenshot 2025-07-02 at 1 21 22 PM

### DIFF
--- a/internal/ui/common/msgs.go
+++ b/internal/ui/common/msgs.go
@@ -18,6 +18,7 @@ type (
 		Output string
 		Err    error
 	}
+	UpdateRevisionsSuccessMsg struct{}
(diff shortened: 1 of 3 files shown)


Gogopex:
thanks! 

>I will merge this even though it doesn't play nicely when experimental_log_batching_enabled = true. I can fix it as a follow up.

Sorry about that! should have been more thorough and read a bit more to catch up!

idursun:
No worries, it is experimental so doesn't really count and I believe I fixed it in a735c26

PR #158, opened 2025-07-04
fix: refresh SelectedFile on ToggleSelect

IvanVergiliev:
When the preview panel is open alongside the details view for a single revision, it displays the currently selected file. Navigating through the file list using up and down arrows changes the preview accordingly. However, pressing the ToggleSelect key moves the cursor in the file list, but it doesn't update the preview panel.

I'm not familiar with the bubbletea library so I mostly followed existing patterns. I tested that the change works in the repo I noticed the issue in.


IvanVergiliev:
Nice, thanks for the extremely quick reaction time!

PR #161, opened 2025-07-04
go mod tidy & go fmt

ilyagr:
Minor maintenance.

PR #162, opened 2025-07-04
ci: run `go fmt` and `go mod tidy` in CI

ilyagr:
~~This is a draft mainly because it depends on https://github.com/idursun/jjui/pull/161.~~

Or you can just merge this instead of merging #161.


idursun:
Thanks! Let’s see how this goes.

PR #165, opened 2025-07-08
feat: add --period/-p CLI option

### LINKED ISSUE #148: FR: command-line argument to set the auto-refresh period
I tried using `jjui` in the `nixpkgs` repo, and it didn't work well because the 4-second auto-refresh interval I configured is too short.

It'd be nice if I could run something like `jjui -n 10` to set a 10-second interval or `jjui -n 0` to disable auto-refresh.

Another option I wondered about would be to allow per-repo configs for jjui (e.g. by reusing jj's per-repo config), but I think a CLI option would be nice regardless.

The `-n` name would conflict with #113. I picked it because the `watch` command calls it `-n` or `--interval`, and I'm used to it. I don't care much what it's called. Random thoughts: perhaps `--interval` is good. `-s` (for "seconds") might work. `-i` seems a bit sketchy; it works, but would be more intuitive for something like "interactive".

ilyagr:
I think `--period` is better than `--interval`, because `-i` sounds more like "interactive". 

Fixes #148

-----

We could still go with `--interval`, or something else if we can come up with something better.

### DIFF
--- a/cmd/jjui/main.go
+++ b/cmd/jjui/main.go
@@ -33,6 +33,7 @@ func getVersion() string {
 
 var (
 	revset     string
+	period     int
 	version    bool
 	editConfig bool
 	help       bool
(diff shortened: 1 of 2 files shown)


idursun (approved):
Thanks!

ilyagr:
Thank you!

PR #171, opened 2025-07-11
feat(git): add UI and test support for 'git push --change <changeId>'

vic:
- Added UI item for 'git push --change <changeId>' when a commit is present.
- Updated tests to verify filtering and selection triggers the correct command.
- Ensured bookmark list expectation is handled in tests.


idursun (approved):
Thank you very much!

PR #182, opened 2025-07-14
feat(exec): Exec key `:` one-shot interactive jj commands.

### LINKED ISSUE #87: Feature request: popup for evaluating arbitrary jj commands
Thanks for this wonderful tool, I use it every day!

I would encourage the developers to consider adding an shell-like window or popup within jjui to allow the execution of arbitrary jj commands.

One of the problems I keep encountering with JJ TUIs is that, if you need to do something specific that isn't supported by the UI, you need exit the TUI to return to the shell where you can run commands normally before reopening the TUI, which can be a bit jarring.

For example, creating bookmarks, duplicating revisions, adding or editing remotes, and other things cannot currently be done in the TUI.

Additionally, especially for newbies, it is sometimes easier to type the jj command as you would in the shell than it is to remember the shortcut key (e.g., it can be easier to type `absorb` vs `abandon` than it is to remember which is `a` and which is `A`).

I am imagining something like vim's command mode, accessed with `:`. Input would be interpreted as arguments to `jj` and any stdout output would be captured and displayed within the interface until esc is pressed.

vic:
This feature is inspired by Vim's `:` key and uses the footer line-editor to obtain input just like we have for `/` (search).

For example, using `: restore -i` will run `jj restore -i` command in interactive mode. jj will use the terminal to show its TUI.

## Context aware placeholders.

We use the same facilities that allow CustomCommands to replace pleaceholders like `$file`, `$change_id`, `$operation_id`.

If the user specifies any of these as arguments to their commands they will be replaced with the context value.

## Exec keys shown in help screen

We have two different exec keys: `:`, `$`.

- `:` 
    As described above, allows running jj interactive commands. These commands can make use of the terminal for TUIs.
   Examples: `: help`, `: squash -i`, etc.

- `$`
  Run interactive shell commands via `$SHELL -c "<input>"`.
  Examples: `$ man jj`, `$ jq . $file | bat --paging always`, `$ htop`

Programs are run interactively with full control of stdio, so users can see all output and errors on the terminal as they'd normally do.  Because of this we don't use status spinner nor error notification on the main UI. However, if the program terminates quickly (less than 5 secs) we assume the program was non interactive (eg, `: version` or `$ ls -la`) and we ask the user for confirmation before closing the terminal and returning back to the main UI.

---

## Try this branch.

To try this branch changes, use:

```shell
nix run github:vic/jjui/exec --refresh
```

----

Closes #87

### DIFF
--- a/internal/ui/exec_process/exec_process.go
+++ b/internal/ui/exec_process/exec_process.go
@@ -1,0 +1,117 @@
+package exec_process
+
+import (
+	"bufio"
+	"io"
+	"os"
+	"os/exec"
+	"strings"
+	"time"
+
+	tea "github.com/charmbracelet/bubbletea"
+	"github.com/idursun/jjui/internal/jj"
+	"github.com/idursun/jjui/internal/ui/common"
+	"github.com/idursun/jjui/internal/ui/context"
+)
+
+func ExecMsgFromLine(prompt string, line string) common.ExecMsg {
+	line = strings.TrimSpace(line)
+	switch prompt {
+	case common.ExecShell.Prompt:
+		return common.ExecMsg{
+			Line: line,
+			Mode: common.ExecShell,
+		}
+	default:
+		return common.ExecMsg{
+			Line: line,
+			Mode: common.ExecJJ,
+		}
+	}
+}
+
+func ExecLine(ctx *context.MainContext, msg common.ExecMsg) tea.Cmd {
+	replacements := ctx.CreateReplacements()
+	switch msg.Mode {
+	case common.ExecJJ:
+		args := strings.Fields(msg.Line)
+		args = jj.TemplatedArgs(args, replacements)
+		return exec_program("jj", args)
+	case common.ExecShell:
+		// user input is run via `$SHELL -c` to support user specifying command lines
+		// that have pipes (eg, to a pager) or redirection.
+		program := os.Getenv("SHELL")
+		if len(program) == 0 {
+			program = "sh"
+		}
+		args := []string{
+			"-c",
+			jj.ShellTemplatedLine(msg.Line, replacements),
+		}
+		return exec_program(program, args)
+	}
+	return nil
+}
+
+// This is different from command_runner.RunInteractiveCommand.
+// This function does not capture any IO. We want all IO to be given to the program.
+//
+// If program terminates in less than 5-secs we ask to press a key to return to JJUI.
+// This is useful for programs that would otherwise terminate quickly and just flash.
+//
+// Since programs are run interactively (without capturing stdio) users have
+// already seen output on the terminal, and we don't use the usual CommandRunning or
+// CommandCompleted machinery we use for background jj processes.
+// However if the program fails we ask the user for confirmation before closing
+// and returning stdio back to jjui.
+func exec_program(program string, args []string) tea.Cmd {
+	p := &process{program: program, args: args}
+	return tea.Exec(p, func(err error) tea.Msg {
+		return common.RefreshMsg{}
+	})
+}
+
+type process struct {
+	program string
+	args    []string
+	stdin   io.Reader
+	stdout  io.Writer
+	stderr  io.Writer
+}
+
+// This is a blocking call.
+func (p *process) Run() error {
+	cmd := exec.Command(p.program, p.args...)
+	cmd.Stdin = p.stdin
+	cmd.Stdout = p.stdout
+	cmd.Stderr = p.stderr
+
+	// If program terminates quickly (most likely non interactive commands),
+	// we ask the user to press a key, so they can at least see the output.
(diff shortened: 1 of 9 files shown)


vic:
Branch is runnable if you want to test it.

~~I still have to fix a test failure (happens in CI, but I'm trying to see why locally all tests are green)~~

internal/ui/context/exec.go
```diff
@@ -0,0 +1,71 @@
...
+	"strings"
+
+	tea "github.com/charmbracelet/bubbletea"
+	"github.com/idursun/jjui/internal/jj"
+	"github.com/idursun/jjui/internal/ui/common"
+)
+
+func ExecMsgFromLine(prompt string, line string) common.ExecMsg {
+	line = strings.TrimSpace(line)
+	switch {
```
vic: `switch prompt` instead of using case booleans.

idursun:
Hey, thanks for working on this!

This is pretty much how I would implement this, with the exception of separating `:` and `|` modes, as it was not clear to me when I was testing it.

I had done some work on this locally but hadn't pushed the branch. I pushed it to the `exp/run-jj-commands` branch now. 

The difference in my implementation is that I used `tea.Exec`, which takes a `tea.ExecCommand` interface implementation. This has the benefit of providing the full shell environment when running the command and handles the coloured output and user input. Maybe we can adapt that approach to your PR? What do you think?

vic:
Sure!, will take a look at `tea.Exec` (and your branch) later today.

vic:
Oh, now I get it, `tea.Exec` let us have an interface that bubbletea will use for yielding stdin/stdout/stderr  so the user can interact normally with the new process with bubbletea out of its way.

Yes I'd like to use that, now that I'm using this branch I just used `$ nix fmt` to run my formatter but see no output at all (all stderr is being captured somewhere I guess), so tea.Exec is definitely handy there. 

My current (past) approach was to re-use the command executor we already had for running `jj` commands.
So I did read our code on how we were doing it and how are custom-commands run. One good thing about this is that I got error handling and nice error reporting for free.

But now  I see the advantage of using tea.Exec  for `:` and `$` (and even in a future pull-request, also for our custom-commands feature)

I will also probably remove `|`, lets see how the `tea.Exec` experience feels. 

So, let me iterate a bit on this, using tea.Exec, and I guess I will also revert some changes (like adding ShellInteractiveCommand to our executor interface) that were just trying to re-use what we already had.

I will also try to impact less our current executor code.

I'm new to Go, so I was unaware of `strings.Fields` will use that instead of `strings.Split`.

Thank you :)

vic:
Hey @idursun,

This branch is much smaller now, and following your advice on using `tea.Exec` proved to be much better and we no longer need `|` to pipe into pager.

This changeset is now more isolated and tries not to touch things we previously had for executing and capturing jj output.

You can test it using 

```
nix run github:vic/jjui/exec --refresh
``` 

Only one thing is missing, that I have not found how to solve:  -its 1:36am over here and I just need some sleep- 

When the process terminates, even when I'm sending a `common.CommandCompletedMsg`, the spinner is still running on the status bar. Maybe you can figure out much better than me what I'm missing.

internal/ui/exec_process/exec_process.go
```diff
@@ -0,0 +1,127 @@
...
+}
+
+// This is different from command_runner.RunInteractiveCommand.
+// This function does not capture any IO. We want all IO to be given to the program.
+// Upon program termination we ask the user to press a key to return to JJUI.
+// This is useful for programs that would otherwise terminate quickly and just flash.
+func exec_program(program string, args []string) tea.Cmd {
+	p := &process{program: program, args: args}
+	return tea.Batch(
+		common.CommandRunning(args),
```
vic: @vic:  CommandRunning should take `program` not only args as we had in the previous patch so that the status line reflect the correct command being run. Otherwise user will see `jj -c whoami` if running `$ whoami`.

idursun:
Thanks! I quickly skimmed through it and it looks good. I will have a proper look when I have the time.

internal/ui/exec_process/exec_process.go
vic: I decided not to use the `CommandRunning`/`CommandCompletedMsg` because I believe they are useful only when we are running commands in the background (as we do for `jj push` for example). But in the case of this feature, programs are already run interactively with full stdio handled to the process, because of that the user has already seen any error produced by the program via stdout/stderr. So I think we dont need to use the status spinner nor error handling.

vic:
This feature is ready now.

I decided not to use the status bar spinner nor error reporting (also added a comment on code about this): 
https://github.com/idursun/jjui/pull/182#discussion_r2211113847

idursun (approved):
Beautiful work! Thank you very much!

PR #185, opened 2025-07-16
test: fix flaky test

idursun:


vic:
got bitten by it: #186

ilyagr:
🤞

ilyagr:
@vic, You can squash an empty commit into your PR to update the committer date and push again if you want to rerun the check. The Git analogue would be `git amend` (I wrote before remembering which project this discussion is in).

vic:
@ilyagr you are right! will try that to trigger the build again. thanks :)

idursun:
I am going to merge this although this is not a definitive fix. It should reduce the flakiness by some amount.

Currently, the test view is closing itself when a `Close` message is received and suffering from a race condition due to the use of `tea.Batch`.

vic:
> Currently, the test view is closing itself when a `Close` message is received and suffering from a race condition due to the use of `tea.Batch`.

Yeah, tea.Batch does not guarantee sequential processing of cmds, I had to [implement proper](https://github.com/idursun/jjui/blob/46264ec3bfb08549a96794a4eab83941f5d78324/internal/ui/ui.go#L56C1-L58C3) [cmd sequencing](https://github.com/idursun/jjui/blob/46264ec3bfb08549a96794a4eab83941f5d78324/internal/ui/leader/leader.go#L59) otherwise keys send by Leader were handled in non-deterministic way. We could move that TakePending out of the leader module and re-use that.

idursun:
I think the use of `tea.Batch` is correct, but the test code isn't handling it well.

PR #193, opened 2025-07-18
fix(revisions): Using `J` should work on uppercase templated change_id

### LINKED ISSUE #191: Moving to Parent in revisions view.
Having the following graph, and starting the top, and using `J`
repeatedly to move to parent revisions, I expected the cursor to move like this:

`W`, `78` (change_id), `K`, `M`, `N`, `YPQP`, `YSZ`

```
○  W vborja@apache.org 1 minute ago C
│  (no description set)
│ ○  R?? vborja@apache.org 1 minute ago A
├─╯  CONFIG_HOME
○  P?? vborja@apache.org 1 minute ago 78
│  (empty) (no description set)
│ ○  R?? vborja@apache.org 1 minute ago E9
│ │  CONFIG_HOME
│ ○  P?? vborja@apache.org 1 minute ago E6
├─╯  (no description set)
○  K vborja@apache.org 1 minute ago context* 5
│  feat(leader): Allow leader entries to specify context requirements.
@  M vborja@apache.org 1 minute ago 77
│  (no description set)
◆  N ibrahim@dursun.cc 21 hours ago main@idursun git_head() 2
│  refactor: unify jj and jjui color unmarshalling logic
◆  YPQP ibrahim@dursun.cc 22 hours ago 462
│  refactor: change batched command order
◆  YSZ vborja@apache.org 22 hours ago DA7
│  fix(exec_shell): Dont replace context values but set env.
```

Master only moves when there's a forkpoint (from `W` to `78`) and not for example from `YPQP` to `YSZ`.

Also tried to fix this at [GetParent](https://github.com/idursun/jjui/blob/main/internal/jj/commands.go#L276) but I'm unable yet to come up with the correct revset expression that works like I expected from `W` to `YSZ`.

Maybe you know what should be the correct expression to use in GetParent.

vic:
This change fixes the `J` move to parent when change_id is not lowercase. 

Having the following graph, and starting the top, moving with `J` repeatedly should work like this:

`W`, `78` (change_id), `K`, `M`, `N`, `YPQP` ...

```
○  W vborja@apache.org 1 minute ago C
│  (no description set)
│ ○  R?? vborja@apache.org 1 minute ago A
├─╯  CONFIG_HOME
○  P?? vborja@apache.org 1 minute ago 78
│  (empty) (no description set)
│ ○  R?? vborja@apache.org 1 minute ago E9
│ │  CONFIG_HOME
│ ○  P?? vborja@apache.org 1 minute ago E6
├─╯  (no description set)
○  K vborja@apache.org 1 minute ago context* 5
│  feat(leader): Allow leader entries to specify context requirements.
@  M vborja@apache.org 1 minute ago 77
│  (no description set)
◆  N ibrahim@dursun.cc 21 hours ago main@idursun git_head() 2
│  refactor: unify jj and jjui color unmarshalling logic
◆  YPQP ibrahim@dursun.cc 22 hours ago 462
│  refactor: change batched command order
◆  YSZ vborja@apache.org 22 hours ago DA7
│  fix(exec_shell): Dont replace context values but set env.
```

Closes #191

### DIFF
--- a/internal/ui/revisions/revisions.go
+++ b/internal/ui/revisions/revisions.go
@@ -3,10 +3,11 @@ package revisions
 import (
 	"bytes"
 	"fmt"
-	"github.com/idursun/jjui/internal/ui/operations/duplicate"
 	"log"
 	"slices"
 	"strings"
+
+	"github.com/idursun/jjui/internal/ui/operations/duplicate"
 
 	"github.com/idursun/jjui/internal/parser"
 	"github.com/idursun/jjui/internal/ui/operations/describe"
@@ -470,11 +471,15 @@ func (m *Model) requestMoreRows(tag uint64) tea.Cmd {
 }
 
 func (m *Model) selectRevision(revision string) int {
+	eqFold := func(other string) bool {
+		return strings.EqualFold(other, revision)
+	}
+
 	idx := slices.IndexFunc(m.rows, func(row parser.Row) bool {
 		if revision == "@" {
 			return row.Commit.IsWorkingCopy
 		}
-		return row.Commit.GetChangeId() == revision || row.Commit.ChangeId == revision || row.Commit.CommitId == revision
+		return eqFold(row.Commit.GetChangeId()) || eqFold(row.Commit.ChangeId) || eqFold(row.Commit.CommitId)
 	})
 	return idx
 }


vic:
From #191:

No elided revisions, revset is .., it makes sense not to jump if the parent is not in rendered in the view. However in my example that is not the case, all parents are visible but we are unable to move to them.

I started a [draft PR](https://github.com/idursun/jjui/pull/193/files#diff-a101ae4c472f500ae5f63f1bc6c871dcab7d5f9d467d32c239d8df334a3f9d06R281) to work on this. Renamed our current GetParent to GetRevisionsParent to make it explicit about possibly taking many revisions as used by squash. And added one simpler version that expects a single revision (used for J on the revisions view).

However it works from W to K, but not from K to M for some weird reason, and works again from M to YPQP but not from YPQP to YSZ hahaha.

idursun:
Oh, I see your commit id template is different than what I had (it's uppercase).

`jjui` parse the coloured `jj log` output and tries its best to figure out the _change id_ and _commit id_. It works fine with the all number commit ids but not with the ones that has capital letters.

Any easy test this idea is turn off uppercasing of commit ids, and it should work as expected. (I mean the old implementation should work as well).

idursun:
Ok, I did some debugging. The problem is that the `GetParent` returned a lowercase commit id but the log contains uppercase commit id.

What we need to fix #191 is to change the [selectRevision](https://github.com/idursun/jjui/blob/f484a38538c7add7230d402ec907b2690f83eb1b/internal/ui/revisions/revisions.go#L470) so that the comparison is case insensitive.

vic:
Haha ok, let me try using case-insensitive `selectRevision` and adapt this PR. Thanks :)

vic:
@idursun works great in all cases I tested even conflicting changes, thank you!

PR is ready now.

idursun (approved):
Thanks!

PR #201, opened 2025-07-22
Dark and light theme support

zerowidth:
This PR adds support for `ui.dark_theme` and `ui.light_theme` to allow for themes in both light and dark mode.

The priority order for theme selection is:

- in light mode, `ui.light_theme` -> `ui.theme` -> default light theme (embedded)
- in dark mode, `ui.dark_theme` -> `ui.theme` -> default dark theme (embedded)


zerowidth:
I could also see themes working the same way UI colors do: if the theme specification is incomplete, start with the built-in theme as the base then merge any overrides. For example, the only diff I have between the default themes is changing the `"selection"` formatting. Let me know if you want me to add that.

idursun:
Hey, thanks for this!

> I could also see themes working the same way UI colors do: if the theme specification is incomplete, start with the built-in theme as the base then merge any overrides.

I was going to suggest this but then reading it a couple of times, I am not sure if we mean the same thing. "working the same way UI colours do" means to me:

```toml
[ui]
theme = "default"
```
is equivalent to

```toml
[ui.theme]
light = "default"
dark = "default"
```

So, then the user can do this:

```toml
[ui.theme]
light = "default_light"
dark = "default_dark"
```

Is that what you meant also?

This way we would be adding this as a new feature on the top of the existing system, not next to it.

zerowidth:
The way I implemented this so far is to preserve the `ui.theme = ...` setting, but allowing specific override for light/dark, e.g. 

```toml
[ui]
light_theme = "my-custom-light-theme"
```

Which would use default dark, custom for light.

---

What I meant by "override the default theme like `ui.colors`", I meant more something like this. Previously I was using:

```toml
[ui]
  highlight_light = "#E6E4D9"
  highlight_dark = "#282726"
```

And now, with this branch, I can do:

```toml
[ui]
  dark_theme = "my_dark"
  light_theme = "my_light"
```

But it feels like kind of too much, given that the _only_ difference in my two themes is

```toml
selected = { bg = "#282726" }
```

And the light equivalent. Everything else is copy/pasted from the default. I'm suggesting that maybe that's _all_ that needs to be in my custom theme files, if I'm happy with everything else.

---

Per my question in discord, I've now updated this PR to also include the embedded config files.

zerowidth:
I've added the default theme with override support. Not only does it make custom themes simpler (if only a few things are overridden), I think this will prevent confusion if/when you add new styled elements. If a user has a custom theme, `jjui` will now use the default color defined internally unless otherwise overridden.

This makes my custom light theme simpler:

```toml
selected = { bg = "#E6E4D9" }
"revisions details selected" = { bg = "#E6E4D9" }
"revset completion selected" = { bg = "#E6E4D9" }
```

idursun:
Oops, sorry for the conflict!

idursun:
Thank you very much for this! 

I had a quick look at the PR and I am fairly happy with the changes and go ahead merge this. However, I still think that the theme value should be a table of `light` and `dark`

```toml
[ui.theme]
light = "my-light-theme"
dark = "my-dark-theme"
```

but with the flexibility that the user can also set it to a single value:

```toml
[ui]
theme = "my-theme"
```
which will resolve to

```toml
[ui.theme]
light = "my-theme"
dark = "my-theme"
```

I can make that change myself after this is merged though. Just wanted to let you know.

zerowidth:
I like the idea of `[ui.theme]`, but I don't think you can have it both ways with a `ui.theme` key and a `ui.theme` table in TOML.

idursun:
> I like the idea of `[ui.theme]`, but I don't think you can have it both ways with a `ui.theme` key and a `ui.theme` table in TOML.

I am already doing it with the colours here: https://github.com/idursun/jjui/blob/095255b4bb87c474618dd1774f9443d0b4bfad18/internal/config/config.go#L39-L77

The following styles are the same.

```toml
"style1" = "red"
"style2" = { fg = "red" }
```

`jj` is doing the same thing with colours too. (Actually that's why I had to do it this way)

zerowidth:
Ahhh ok, because it's a flexible deserialization and it's _either_ a value _or_ a table, but not both. Makes sense!

 I can't get back to this PR until tonight, but I'd be happy to make that change if you don't get there first. Thank you!

idursun (approved):
This is good to go!

I don't know why nix check is failing, as none of the dependencies changed in this PR. I can have a look at it.

Also, there was a new conflict, which I believe I fixed, but still says cannot be merged due to conflicts. I don't know what's going on there.

PR #203, opened 2025-07-22
feat(fuzzy_files): add ctrl+t fuzzy-file search.

vic:
This feature implements `ctrl+t` to locate files and changes affecting them using fuzzy search.

It allows to quickly search over all files that exist on a revision. 

This is not the same as the file list we already show in details view (which shows only the files changed in a given revision).

---

Add fuzzy file search feature with Ctrl+T keybinding that allows users to quickly search and navigate to files within a specific revision.

- Add FileSearch key binding (Ctrl+T) to key configuration
- fuzzy_search package is generic enough to be used for other fzf-like
  fuzzy completions, it features refined searching and default rendering.
- fuzzy_files UI component implements specifics for file searching.
- Add FilesInRevision command to list files in a specific revision
- Integrate fuzzy file search with main UI and status bar
- Move sahilm/fuzzy from indirect to direct dependency in go.mod
- Add preview support and revisions targeting for fuzzy search
- Include file search in help page documentation

### DIFF
--- a/internal/config/keys.go
+++ b/internal/config/keys.go
@@ -32,6 +32,7 @@ var DefaultKeyMappings = KeyMappings[keys]{
 	Revset:            []string{"L"},
 	ExecJJ:            []string{":"},
 	ExecShell:         []string{"$"},
+	FileSearch:        []string{"ctrl+t"},
 	QuickSearch:       []string{"/"},
 	QuickSearchCycle:  []string{"'"},
 	CustomCommands:    []string{"x"},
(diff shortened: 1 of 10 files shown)


vic:
PR ready :)

vic:
thanks for the feedback. I'll be working on them later today and will probably get back to you.

I will also split this change in two, so that #210 does not depend directly on this PR of fuzzy_files #203  but only on the fuzzy_search module itself, if that makes easier for #210 to be merged first.

vic:
> * When I press `j`, `k` commands, I can see that revisions cursor is moving up and down. While editing, all the keys should be handled by the input.

fixed this one.

vic:
> The view is pushing the revisions and preview windows upwards.

Yes, currently our fuzzy view is part of the status view so if it expands to show candidates, the left and preview content is shortened.

I'm investigating how stacking works, but I'll also read how our notifcation popups work, because I'm not really sure I'd like to use the stacking window for this.

The stacked view is too far from the input, and completions should be near where I am typing, that's actually why I made matches be presented with the best match being lower, immediately near to the input, not having to move my eyes from where I'm typing to where the best suggestion is shown is important.

So I'll investigate if I can use something like the floating popups we have.

vic:
> polluting the revset history

fixed

vic:
> * I'd prefer it only to update the revset when the user selects a file.

done, this the default mode when you enter `ctrl+t`.
if you press `ctrl+t` again it toggles live revset and change preview.

vic:
> * The view is pushing the revisions and preview windows upwards.

I'll address this one now (the only one remaining of the issues you mention)

vic:
Ok looks like I'll be trying to use `screen.Stacked` just like we have for flash messages, but I'll create another one and position it near the status bar input.

I just wont be using `UI Model.stacked` which is used for help and git windows I guess and was the one I was referring as too far from the input.

vic:
> * This might look better if it was stacked on the top of the main view. Stacking happens in `ui.go`

done ! It looks much much better stacking on top :) Thanks for your suggestion!

vic:
PR ready (again)

idursun (approved):
Brilliant! Thanks!

PR #204, opened 2025-07-22
fix(details): Quote file paths when changing revset from details.

vic:
This fixes an error when changing the revset from details view when the file path starts with a dot.

.github/workflows/foo

Without this change, the revset sees the first dot as a syntax error (trying to call a function). To prevent this, we quote the full path since maybe other paths could be interpreted as jj expressions unless we quote them as we do now.

Noticed about this when using #203.


idursun (approved):
Thanks!

PR #206, opened 2025-07-22
feat: add command outputs as (auto-expiring) flash messages

idursun:
Previously, revset errors were displayed under the revset component and similarly command errors were displayed under the status line. This not only looks out of place but also complicates the widget implementations unnecessarily. 

I have introduced a flash message view, which displays success and error messages at the bottom right corner. Success messages expire after 4 seconds, but error messages are sticky and they are dismissed by pressing `esc`.

Multiple messages stack on top of each other.

image

image

image


vic:
I'm in love with this feature <3 thank you!

PR #208, opened 2025-07-22
WIP: load/write history for search/exec editor

vic:
As first part of progress on #202

PR #210, opened 2025-07-23
feat(exec): add fuzzy search on command history

### LINKED ISSUE #202: FR: History for commands in Exec `:` and `$`
The "Exec" command prompts keep a history of past commands available e.g. via arrow up or arrow down or chosen from a list with Ctrl-r.

Implementing full history functionality might not warrant the improved convenience, putting it up to see your thoughts on it nonetheless.

A minimal implementation might already provide most of the value, but maybe there are 3rd party history APIs that could be used?

vic:
Closes #202
Depends On #208 ~and #203~ 

--

The fuzzy input allows users to quickly find and select from command history 
using substring matching with highlighted results.

Added a fuzzy_input component that works along textinput, using the
same keybindings and automatically showing suggestions as textinput did.

When a fuzzy_input is initialized with a textinput, it disables the
textinput suggestions and replaces them with an interactive
fuzzy search interface.

Suggestions are accepted using the same keys as textinput, via TAB key.
Supports moving the selected suggestion using same keys as textinput.

No need to use `ctrl+r` since suggestions are shown automatically.

Tested locally doing the following to load about 9934 history entries.

```
history >> ~/.cache/jjui/history/exec_sh
```

You can run this branch as usual, using nix:

```
nix run github:vic/jjui/input --refresh
```


fleimgruber:
I am not at my private NixOS system. So I wanted to see how I can run it with go directly from sources, how would I do that? I tried with `go run cmd/jjui/main.go` from the PR branch but I can't see the feature there.

*edit*: I can see the feature working now, arrows up and down work as well as fuzzy matching - looks nice!

vic:
No longer depends on #203

idursun (approved):
Thanks! Worked fine for me.

I haven't checked the stylings but I guess I will do it in another pass myself.

PR #215, opened 2025-07-24
feat(details): show conflicted files

### LINKED ISSUE #209: FR: Make it easy to identify files having conflicts on details view.
When a revision has conflicts, and you open the details view, it would be nice to have a visible way to know which files have conflict.

This can help quickly identifying which file you should edit, via a leader key or something without having to shell-out to `: status` to find out.

idursun:
Adds conflict marker to the file list in the details view. I had to use a template to get this information which is only available after jj v0.26+.

I have tested it manually, and works fine. I will add some tests for this when I get to resolving the flakiness in the details tests suite.

image

fixes #209

### DIFF
--- a/README.md
+++ b/README.md
@@ -147,7 +147,7 @@ You can download pre-built binaries from the [releases](https://github.com/idurs
 
 ## Compatibility
 
-Minimum supported `jj` version is **v0.21**+.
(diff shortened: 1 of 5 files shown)


vic:
Looks nice! thank you!

Issue #184, opened 2025-07-16
Feature Idea: Edit file from Details


maralorn:
More an idea, than a request.

It would be very convenient if one could go to a file in the details list of change, press `e` and be dropped into a text editor to edit that file.

This would be especially fancy if it would edit the file in the selected change.

vic:
You can now, using [two](https://github.com/idursun/jjui/pull/182) [new](https://github.com/idursun/jjui/pull/179) features that I've contributed and are now merged in master:

you can define a [leader](https://github.com/idursun/jjui/wiki/Leader-Key) key that sends the following keys:

```
[leader.e]
help = "Edit current file"
send = [ "$", "jj edit $change_id && vim $file", "enter" ]
```

vic:
Have you tried it ? I'm using it on my own configs and it is neat, thanks for the idea ! 

If it works for you feel free to close this issue.

maralorn:
Well, this seems relatively orthogonal to the leader key. And it was already possible to configure custom commands.

The problem with your example is that it changes the change currently being edited. I would prefer to only modify the file at the change without influencing the worktree.

But yeah, your command looks useful, I will try it out.

vic:
our custom-commands are for executing jj commands defined at user config. it was not possible to invoke an editor from them unless you relied on `jj util exec` (but they are experimental and might be removed). our exec `:` and `$` facilities basically allow you to run anything as a convenience without leaving jjui. for running short commands it's great.

our exec combined with leader, let you have custom workflows (when I'm hacking on jjui I have a key to run go tests on a particular file using a modified version of the leader key for editing), so pretty much anything your heart desires (either good or bad) is possible haha.

why would you edit a file without making its revision also editable. (I'm really curious, is that any good ? maybe I'm missing something). I know there is a project jj-run that lets you run arbitrary commands on a temporary workspace, it would let you edit the file without moving your current @. maybe you might find it useful.

vic:
>  invoke an editor from them unless you relied on `jj util exec` (but they are experimental and might be removed). 

even then, the way custom-commands work, they capture stdio since it uses the same code path that jjui uses for running jj commands (capturing stderr). our exec features however yield complete stdio control to whatever command you type. for interactive commands like running an editor or a code formatter from within jjui that's an important difference.

maralorn:
Okay, that sounds actually sweet. And I will look into jj-run.

[...]

PR #219, opened 2025-07-25
feature(details): Allow to edit selected files from details view.

vic:
This is useful when solving conflicts, opening the details view now shows files with conflicts, you can select them and open an editor.

See also #184.

### DIFF
--- a/internal/config/default/config.toml
+++ b/internal/config/default/config.toml
@@ -58,6 +58,7 @@ limit = 0
     restore = ["r"]
     absorb = ["A"]
     diff = ["d"]
+    edit = ["alt+e"]
     select = ["m", " "]
     revisions_changing_file = ["*"]
   [keys.preview]
--- a/internal/config/keys.go
+++ b/internal/config/keys.go
@@ -67,6 +67,7 @@ func Convert(m KeyMappings[keys]) KeyMappings[key.Binding] {
 			Restore:               key.NewBinding(key.WithKeys(m.Details.Restore...), key.WithHelp(JoinKeys(m.Details.Restore), "restore")),
 			Absorb:                key.NewBinding(key.WithKeys(m.Details.Absorb...), key.WithHelp(JoinKeys(m.Details.Absorb), "absorb")),
 			Diff:                  key.NewBinding(key.WithKeys(m.Details.Diff...), key.WithHelp(JoinKeys(m.Details.Diff), "diff")),
+			Edit:                  key.NewBinding(key.WithKeys(m.Details.Edit...), key.WithHelp(JoinKeys(m.Details.Edit), "edit files in revision")),
 			ToggleSelect:          key.NewBinding(key.WithKeys(m.Details.ToggleSelect...), key.WithHelp(JoinKeys(m.Details.ToggleSelect), "details toggle select")),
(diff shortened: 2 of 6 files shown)


idursun:
This is nice, but also comes with the false assumption that the file is going to be edited as it appeared in the selected revision. This might be confusing for users even though for the initiated there's not a way to do that with a single command in `jj`. It can even more confusing if the user tries to edit a file which has been renamed or deleted in a later revision. I believe it will show an empty file.

It's fine as it is not listed in the details view's shortcut list. However, you have already implemented a similar feature in the fuzzy files, that the user can edit the file that they have selected, so this one seems a little bit extra and out of place. What do you think?

vic:
I think I understand your points. I'm currently [using a leader key](https://github.com/idursun/jjui/issues/184#issuecomment-3081972709) to open files from jjui details. Having a leader key is handy, and mine *does moves* to the revision first before editing the file. But now that I'm able to see conflicts in revisions I wanted to open my editor with many selected files, using my leader key here is not convenient (thus this PR) or otherwise I could just invoke my `\e` leader key for each file as I already do, but I wanted to open many selected files at one, that's why I implemented this.

So one thing we could do in this PR is to move `@` to the revision on which the file is expected to be edited.

If so, I'd remove the `e` keybinding, leaving only `alt+e` as to make it a more *conscious* keystroke, because using `alt+e` would make the revision current and edit the file, effectively editing the file in that revision. 

Having only `alt+e` makes it more consistent with #220 where `alt+e` is used to open a file from a revision.

Having a single `alt+e` consistent command for editing files is better because it could map mentally to:  "open the file in the revision it belongs" everywhere in the interface.

Regarding deleted files, and if we implement "make revision current before edit file", using `alt+e` on an removed file would just open an empty buffer, because you'll be moved to the revision where the file no longer exists but you open the path in your editor which is naturally empty.

vic:
PR updated.

vic:
Now there's another issue with `alt+e` edit file from revision. If people try to edit a file on an immutable commit, the editor will open but jj actually denied to "edit the revision" (making it `@`). but you wont notice this (I did not) until you get back to the jjui interface. This is because this current impelmentation just batches `EditRevision` and `EditFile` but the second one has no way to know the first cmd was sucessfully executed.

vic:
But maybe we can live with that since people are trying to edit files on immutable revisions explicitly. Nevertheless, I believe we should not open the editor unless the revision is mutable.

vic:
Maybe if we had a `$files` (plural) placeholder that expanded to an space separated list of the files currently selected from details, I could continue using my `\e` leader key. Having the edit file functionality might be useful for other people as well.

If this PR makes no sense, and you'd prefer to have a  `$files` context value that I could use with my leader key, I can also send another PR for that. 

If you do see value on this, and want me to work on not trying to edit files on immutable revisions, tell me. otherwise the PR is ready for review.

idursun:
> Maybe if we had a $files (plural) placeholder that expanded to an space separated list of the files currently selected from details, I could continue using my \e leader key.

Exactly. I think an approach like this is the solution to the problem. I'd suggest supporting spread syntax like `$...file` which would work with this, but also similarly would work with `$...change_id` opening up many possibilities (e.g. `$...bookmark`, etc) without breaking the existing behaviour.

vic:
Will open another issue to talk about implementing spread placeholders.

I'm closing this for now.

PR #220, opened 2025-07-25
enhance fuzzy-file live preview. add contextual help and open file in editor

vic:
The fuzzy file search now provides much better experience for real-time feedback, improved keybidning discoverability and edit file capabilities.

- Live preview mode now works with debounced revset updates to prevent excessive UI refreshes.

- when the editor on status line is active it is now able to show contextual help or keybindings.

- we now show keybindings available while file finder is active. this allows better key discoverability, since these keys are part of the editing mode and not global to the ui.

- re-use our GetDefaultEditor function and ExecShell to allow the user directly edit by using alt+e while on file finder.

Will open another PR to move hardcoded keybindings to our default toml file.

### DIFF
--- a/internal/ui/fuzzy_files/fuzzy_files.go
+++ b/internal/ui/fuzzy_files/fuzzy_files.go
@@ -4,7 +4,9 @@ import (
 	"fmt"
 	"strconv"
 	"strings"
+	"time"
 
+	"github.com/charmbracelet/bubbles/help"
 	"github.com/charmbracelet/bubbles/key"
 	"github.com/charmbracelet/bubbles/textinput"
 	tea "github.com/charmbracelet/bubbletea"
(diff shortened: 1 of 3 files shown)


internal/ui/fuzzy_files/fuzzy_files.go
```diff
@@ -83,6 +101,7 @@ var inputKm = textinput.DefaultKeyMap
 var up = key.NewBinding(key.WithKeys("up"))
 var down = key.NewBinding(key.WithKeys("down"))
 var enter = key.NewBinding(key.WithKeys("enter"))
+var edit = key.NewBinding(key.WithKeys("alt+e"))
```
idursun, line 104: I know this will go away. You can actually pass the help text to the key bindings

```go
var up = key.NewBinding(key.WithKeys("up"), key.WithHelp("up", "move up in the list"))
var down = key.NewBinding(key.WithKeys("down"), key.WithHelp("down", "move down in the list"))
var enter = key.NewBinding(key.WithKeys("enter"), key.WithHelp("enter", "select file"))
var edit = key.NewBinding(key.WithKeys("alt+e"), key.WithHelp("alt+e", "edit file"))
```

Issue #67, opened 2025-05-01
Add ability to enter custom prompts before running custom commands


idursun:
I think that the only thing missing for me is being able to query the user for custom input to these custom commands. E.g., I would like to create a custom command to create a new bookmark for the current revision. The command should support `bookmark set --revision $change_id $inputs.bookmark_name` where `$prompts` is an array of user inputs specified in the custom command, and the `.bookmark_name` is the specific user input with the value inserted by the user when invoking the custom command through an interactive text field. Something like:
```toml
[custom_commands]
"new bookmark" = { 
  key = ["ctrl+b"], 
  args = ["bookmark", "set", "--revision", "$change_id", "$prompts.bookmark_name"],
  prompts = [ {id = bookmark_name, type = "text", title = "Name of the new bookmark"} ]
}
```
The `prompts.type` could be selecting from a menu (with filtering support), inputting a custom text, ...

Similarly to what [Lazygit does with prompts](https://github.com/jesseduffield/lazygit/blob/master/docs/Custom_Command_Keybindings.md):
```yaml
  - key: 'n'
    context: 'localBranches'
    prompts:
      - type: 'menu'
        title: 'What kind of branch is it?'
        key: 'BranchType'
        options:
          - name: 'feature'
            description: 'a feature branch'
            value: 'feature'
          - name: 'hotfix'
            description: 'a hotfix branch'
            value: 'hotfix'
          - name: 'release'
            description: 'a release branch'
            value: 'release'
      - type: 'input'
        title: 'What is the new branch name?'
        key: 'BranchName'
        initialValue: ''
    command: "git flow {{.Form.BranchType}} start {{.Form.BranchName}}"
    loadingText: 'Creating branch'
```

_Originally posted by @Adda0 in https://github.com/idursun/jjui/discussions/63#discussioncomment-12988811_

vic:
Since I've been recently digging in the source code, I'd like to work on this. However, before doing so, I will be thinking and getting back to you on a proposed design for this. (designing is part of working on it)

I understand the use-case for having prompts and the value they add for custom-commands.

However, given that custom-commands are user defined, and we can't possibly control what they do, why not just let custom-commands be user-configured names (and keys) for command-lines that can be run via our new  `ExecShell`.

[...]

PR #222, opened 2025-07-25
Add support for executing jj aliases.

### LINKED ISSUE #211: Run aliases from ui
Seems like it would be nice to be able to run entries from the `[aliases]` section of the configuration. For example:

```toml
[aliases]
bm = ["bookmark", "set", "main", "--revision=@"]
sync = [
    "util",
    "exec",
    "--",
    "bash",
    "-c",
    'jj git fetch && jj rebase -d "trunk()"',
]
```

These are of course operations that can be done in the ui already, but to be able to run any of the aliases seems like it would be nice.

theherk:
This pops up a menu of aliases already existing in jj config, and supports parameter input.

Fixes #211

### DIFF
--- a/README.md
+++ b/README.md
@@ -60,6 +60,16 @@ For detailed information, see [Preview](https://github.com/idursun/jjui/wiki/Pre
 
 ![GIF](https://github.com/idursun/jjui/wiki/gifs/jjui_preview.gif)
 
+### Aliases
+You can execute any JJ alias defined in your configuration by pressing `X`. This opens a menu showing all available aliases from your `jj config`. The menu displays:
+- Alias names and their corresponding commands
+- Parameter requirements (e.g., "needs 1 parameter(s)")
+- Inline parameter input for aliases that use positional parameters like `$1`, `$2`, etc.
+
+When you select an alias that requires parameters, the interface shows an input field below the menu. For aliases with multiple parameters, you'll be prompted for each one sequentially (e.g., "param1: (1/2)", "param2: (2/2)", etc.). Aliases are displayed in alphabetical order for easy navigation.
+
+For example, if you have an alias `co = ["new", "$1"]`, selecting it will show an input field where you can type the parameter that will replace `$1` in the command.
+
 Additionally,
 * View the diff of a revision by pressing `d`.
 * Edit the description of a revision by pressing `D`
@@ -71,6 +81,7 @@ Additionally,
 * Git _push_/_fetch_ by pressing `g`
 * Undo the last change by pressing `u`
 * Show evolog of a revision by pressing `v`
+* Execute JJ aliases by pressing `X`
 
 ## Configuration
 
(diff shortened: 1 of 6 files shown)


README.md
```diff
@@ -60,6 +60,16 @@ For detailed information, see [Preview](https://github.com/idursun/jjui/wiki/Pre
...
 
+### Aliases
+You can execute any JJ alias defined in your configuration by pressing `X`. This opens a menu showing all available aliases from your `jj config`. The menu displays:
+- Alias names and their corresponding commands
+- Parameter requirements (e.g., "needs 1 parameter(s)")
+- Inline parameter input for aliases that use positional parameters like `$1`, `$2`, etc.
+
+When you select an alias that requires parameters, the interface shows an input field below the menu. For aliases with multiple parameters, you'll be prompted for each one sequentially (e.g., "param1: (1/2)", "param2: (2/2)", etc.). Aliases are displayed in alphabetical order for easy navigation.
+
+For example, if you have an alias `co = ["new", "$1"]`, selecting it will show an input field where you can type the parameter that will replace `$1` in the command.
```
vic, line 71: Since `co` is a `jj` alias, how is `jj co` useful on the command line ? because it expands to `jj new $1` on the system shell execution outside of jjui ?

I'm asking because jj aliases just execute whatever you define on them, `$1` is not handled specially AFAIK. and this makes `co`, an alias made specifically for jjui (with this feature). This defeats the purpose of "re-using aliases from jj in jjui" (as originally mentioned in #211), since now `co` is useful in `jjui` but no in `jj`, _and_ you are now configuring tailor-made things for jjui in jj's configuration file.


For reviewers:

I believe this is basically implementing `custom-commands` again but from jj aliases. where we also have proposals for custom prompts (#67). 
for context, see [#211 comment](https://github.com/idursun/jjui/issues/211#issuecomment-3109740001) for loading existing jj aliases as custom-commands.

vic: I do see value on your work and on being able to invoke jj aliases, it makes sense to me. jj aliases exist because they make people's life easier while executing via the system shell (or via our `:` exec).

I'm just trying not to have other window for user defined commands, and maybe we can re-use your work as the start for prompts (for custom-command) if we unified them with jj aliases. Just my opinion.

Thanks again for working on this ;)

theherk: `co` is simply a bad, contrived example for the reason you point out. It was just one of a few I used to ensure I could handle parameters. My real targets didn't use parameters. I'll update to reflect this later today. 

On the second topic, I considered unifying this with custom command or using in conjunction with `:`, but thought it was a bit cleaner not to. Extra window vs. extra complexity in those interfaces.

theherk: Actually, maybe you can give an example of one of your aliases that uses parameters which would serve as a useful example for documentation. For me, the alias I use that I wanted to run from here was:

```toml
[aliases]
sync = [
    "util",
    "exec",
    "--",
    "bash",
    "-c",
    'jj git fetch && jj rebase -d "trunk()"',
]
```

Though there is probably a better way to do that, too. Nevertheless, I use that with jj cli.

theherk:
Rejecting: not in line with project goals or design guidelines.

PR #225, opened 2025-07-27
fix(commands): don't create snapshots when only querying current state.

vic:
Use `--ignore-working-copy` on non-mutable commands to prevent creating snapshots when we only want to obtain data from jj store.

### DIFF
--- a/internal/jj/commands.go
+++ b/internal/jj/commands.go
@@ -16,11 +16,11 @@ const (
 type CommandArgs []string
 
 func ConfigListAll() CommandArgs {
-	return []string{"config", "list", "--color", "never", "--include-defaults"}
+	return []string{"config", "list", "--color", "never", "--include-defaults", "--ignore-working-copy"}
 }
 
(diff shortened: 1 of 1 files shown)


idursun (approved):
Nice! Thanks!

internal/jj/commands.go
```diff
@@ -16,11 +16,11 @@ const (
 type CommandArgs []string
 
 func ConfigListAll() CommandArgs {
-	return []string{"config", "list", "--color", "never", "--include-defaults"}
+	return []string{"config", "list", "--color", "never", "--include-defaults", "--ignore-working-copy"}
 }
 
 func Log(revset string, limit int) CommandArgs {
-	args := []string{"log", "--color", "always", "--quiet"}
+	args := []string{"log", "--color", "always", "--quiet", "--ignore-working-copy"}
```
idursun, line 23: @vic I had to revert this part because we want to snapshot when `log` is called. Otherwise, the revisions that have changes in it show as empty on start up, or when revset is changed.

PR #226, opened 2025-07-27
feat(revisions): Ace jump to revision. `f` key on revisions view.

vic:
This feature implements a simple vim like ace-jump to quickly move to revisions (`revisions.rows`) from the current revset.

Pressing `f` will highlight the first character of each change_id (only non-hidden, non-conflicting, and non-root) or each commit_id.

Wile ace-jump is enabled a temporary case-insensitive keymap is created that matches the next available keys from revisions ids. When an unambiguous revision is selected via its shortest id, the cursor is moved to that revision row.

This allows to have much faster movements than currently using `/` search and `'`

And also opens the door for a future PR to explore on making `/` work with fuzzy search (to match on full-description and not only on the content of the log being displayed) while preserving fast movements on revisions with `f`

### DIFF
--- a/internal/ui/graph/default_row_iterator.go
+++ b/internal/ui/graph/default_row_iterator.go
@@ -15,6 +15,7 @@ import (
 
 type DefaultRowIterator struct {
 	SearchText    string
+	AceJumpPrefix *string
 	Selections    map[string]bool
 	Op            operations.Operation
 	Width         int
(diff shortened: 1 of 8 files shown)


vic:
~Fixing a panic when a bookmark is set on the target revision.~  was missing a return.
but you can try it if you want to know how it feels like.

vic:
I still have to fix a bug where the `ace` bookmark is somehow identified as a commit_id is being highlighted, dunno why. but will do that on monday, it's 4am here and have to get some sleep, ~tomorrow~ today is baseball day :D

image

vic:
Quick question: is there a way I can get only the `revision.rows` that are currently being displayed on screen ? (not all the rows we have in memory, since not all of them are visible, it makes no sense to have key bindings for the ones that are not being shown)

internal/ui/graph/default_row_iterator.go
```diff
@@ -132,10 +133,20 @@ func (s *DefaultRowIterator) Render(r io.Writer) {
 				style = style.Inherit(s.textStyle)
 			}
 
+			isAce := func() bool {
+				return s.AceJumpPrefix != nil &&
+					(segment.Text == row.Commit.ChangeId || segment.Text == row.Commit.CommitId) &&
```
vic, line 138: > I still have to fix a bug where the `ace` bookmark is somehow identified as a commit_id

PR is ready, but I guess I need help with this, because it seems  we have a bug where sometimes a bookmark is identified as a commit_id. I'm checking here that the segment text is either the row changeId or the commitId, but somehow `ace` (a bookmark -- see my screenshot) is sometimes (not always) identified as a commit id and this condition evaluates to true, making ace jumping into the bookmark instead of the real commit id.

idursun: Ah, yes that might happen. Let me have a look.

vic:
I'm marking this PR as ready for review/merge.

We can solve the "bookmark as commit_id" in another PR since it is not specific to changes here. And even in the presence of the bug this PR seems to work well. EDIT: #228 

I will also work on generating shortcuts only for the revisions visible on screen in another PR.
Still have to study a bit more the code to know how to implement it, because even when AceJump on `..` complete history does work nice and feels responsive, shortcuts are enabled even for revisions out of screen and jumping there seems a bit weird.

idursun:
> Still have to study a bit more the code to know how to implement it, because even when AceJump on .. complete history does work nice and feels responsive, shortcuts are enabled even for revisions out of screen and jumping there seems a bit weird.

The magic happens here: https://github.com/idursun/jjui/blob/499fa83c51eb91752b3bd334d3229374e3e5424e/internal/ui/graph/renderer.go#L108-L132

It currently tracks only the last rendered row index, but we can add the first rendered row id too.

PR #232, opened 2025-07-29
feat(revisions): Move to revision parent when toggle selection.

vic:
Automatically moves the cursor to the revision parent (on same branch) when a revision selection is toggled.

This is a QOL improvement when working on multiple revisions as you no longer have to move and only press space.

Implements the proposal from [#221 comment](https://github.com/idursun/jjui/discussions/221#discussioncomment-13898398) and makes moving-on-selection consistent on both revisions and files.


idursun (approved):
LGTM!

PR #234, opened 2025-07-29
fix(preview): Limit preview resize to range 5% up to 95%.

vic:
This fixes two bugs:

- The preview resize messages were being handled even if the preview window was hidden. This caused strange problems if ctrl+h or ctrl+l were hit while preview was hidden.

  We now only handle resize messages if preview is actually shown. Because these commands effects are intended to be immediatly visible to the user.

- There was no limit to scaling up and when we hit more than 100% the screen was messed up and preview window was no longer able to be resized.


idursun (approved):
Thanks!

PR #236, opened 2025-07-29
feat(evolog): Restore descendants from selected evolog revision.

vic:
Workflow: mark a target revision as selected using ` ` (space). Enter the evolog on source revision using `v`, when hovering the desired source, press `R`.

The following command will be run:

```
jj restore --from SOURCE --into TARGET --restore-descendants
```

See #235


idursun:
Hey, thanks!

I gave it a quick go, and it works. However, first thing I noticed was that I had a custom command bound to `R` and it was invoked instead of the evolog's `R` key. 

I'd suggest the restore keybinding for the evolog to be `r` (which is the same in details view) so they are consistent.

Most of the time user might not know if they want to restore evolog, but when they decide to restore, then they have to close the evolog and choose the target and come back to evolog, choose the entry they want again and then restore, which is a lot of work.

Instead of starting with a selection, let to the user to open the evolog, and hit `r` if they want to restore, and then give them the option select the target as we do in the _rebase_, _squash_ operations. I think this would be ideal flow as showing the restore option regardless of a selection makes it discoverable and also giving the ability to choose the target after they want to restore makes it flexible.

What do you think?

nickchomey:
I strongly agree with the suggested change

idursun:
@vic you are working on so many things at the moment, let me know if the proposed change is distraction to the other work you are doing at the moment. I am happy to implement it on this PR.

vic:
changes sound good to me, and they sound easy when you mention them, haha. I have no idea how to do that option to interactively select the target as we do in rebase/squash. I just implemented it using checked revision because that's how I imagined to have something working asap and see if I could have a prototype of it quickly. If you want to continue this PR, please do so. :)

vic:
> I am happy to implement it on this PR.

Yes please.

arnold-c:
This is all great. Thanks @vic for pulling this together so quickly. The suggested changes align with my proposed workflow, so that works well for me!

idursun:
@vic I think this is now ready to be reviewed.

vic (commented):
LGTM

PR #238, opened 2025-07-29
feat(preview): Allow showing preview at bottom (config and toggle-key)

### LINKED ISSUE #93: Optionally arrange commit graph and preview vertically
When jj is configured to use a diff tool such as https://github.com/Wilfred/difftastic that defaults to side-by-side diff views to use horizontal screen estate, jjui preview pane horizontally aligned with the commit graph does not show the right-hand side of the diff view.

In such cases an option would be to configure the preview pane to be vertically aligned with the commit graph.

I was using lazyjj until today which provides a layout configuration setting to this effect: https://github.com/Cretezy/lazyjj?tab=readme-ov-file#configuration.

vic:
The option `preview.show_at_bottom` (default: false) and the `P` key
allow moving the preview window to the bottom instead of being shown to
the side.

Fixes #93

### DIFF
--- a/internal/config/config.go
+++ b/internal/config/config.go
@@ -120,6 +120,7 @@ type PreviewConfig struct {
 	OplogCommand             []string `toml:"oplog_command"`
 	FileCommand              []string `toml:"file_command"`
 	ShowAtStart              bool     `toml:"show_at_start"`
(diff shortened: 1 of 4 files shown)


idursun (approved):
so good 😍

PR #246, opened 2025-07-31
feat(placeholders): Add $checked_files and $checked_commit_ids.

### LINKED ISSUE #223: WIP: Implementing spread placeholders.
As described in #219 it would be useful to have placeholders with multiple values, particularly for multiple selected files, but we can also have multiple selected revisions on the UI, so having spread placeholders for them allows user defined custom-commands and leader keys to implement things that operate in multiple files/revisions, etc.

For context, #219 was about opening multiple files from the details view and originally inspired by #184. 

This issue is for trying to come up with a syntax that is useful in both custom-commands and leader keys, since they both currently use the `$placeholder` syntax.

On custom-commands, `$placeholder`s are just replaced on the custom-command arguments string. However in exec (both `jj` and `sh`) they are actually environment variables (`placeholder=value`) set for the underlying process before execution.

So, for custom-commands any spread syntax we choose `$...placeholder` (as originally proposed by @idursun) is fine since we simply make string replacement. 

However since shell utilities can be user defined programs (as to avoid having long scripts in jjui config) they can access their contextual placeholders via the process environment: `$file`, `$revset`, etc. 

So, instead of `$...placeholder`, I'm proposing we use `$placeholder_list`, like `$file_list` or `$change_id_list`. They will be substituted (in the case of custom-commands) or set in the environment (for jj and sh exec) as space separated values. space is a commonly used in shells as field separator `IFS`.

However, it is also common for filenames to include spaces, so we either quote files when they are part of `$file_list` or use another separator. 

I'll start implementing this `$file_list` for selected files from the details view. But I'd like some feedback about my previous notes.

vic:
This patch allows custom-commands and exec-commands to have access to user selected revisions and files. By "selected" we mean, marked as _checked_  (✓)  using space ` ` on the UI.

You can test this branch in two ways:

- Select some revisions from the UI and then execute a command like: `$ echo $checked_commit_ids`

- Select some files from the details UI and execute: NOTE: expansion is performed by underlying $SHELL. This example assumes bash:

   `$ ls -la ${checked_files[@]}`

See [#223 comment](https://github.com/idursun/jjui/issues/223#issuecomment-3134918996)

Closes #223

### DIFF
--- a/internal/jj/commands.go
+++ b/internal/jj/commands.go
@@ -12,6 +12,17 @@ const (
 	FilePlaceholder        = "$file"
 	OperationIdPlaceholder = "$operation_id"
 	RevsetPlaceholder      = "$revset"
+
+	// user checked file names, separated by `\t` tab.
+	// tab is a lot less common than spaces on filenames,
+	// and is also part of shell's IFS separator.
+	// this allows programs like `ls -l ${checked_files[@]}`
+	CheckedFilesPlaceholder = "$checked_files"
+
+	// user checked commit ids, separated by `|`.
+	// the reason is user can use checked commits as revsets
+	// given to jj commands.
+	CheckedCommitIdsPlaceholder = "$checked_commit_ids"
 )
 
 type CommandArgs []string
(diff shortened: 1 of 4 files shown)


idursun (approved):
Thanks.

I believe this is just an interim step until we have go templates for custom commands implemented.

PR #247, opened 2025-07-31
feat(revert): Revert operation (`R`) with interactive target selection

vic:
Code is mostly based on `rebase` operation, but simplified a bit since  revert always has revisions as source.


idursun (approved):
Thanks!

PR #251, opened 2025-08-01
Experimental lane tracing support

idursun:
This PR adds highly experimental support for highlighting lanes based on selected revision and dims the out of lane revisions.

It's currently behind a configuration value, you can enable it like the following.

```toml
[ui.tracer]
enabled = true
```

image

image


vic:
Will try this later today, thanks!

vic:
Hey, I know this is a work in progress, and still marked as draft.

Just wanted to share some issues that appear even if tracing is disabled. Like I seem to have underlined spaces near the lane lines.

---
With tracing enabled
Screenshot_20250801_133023

---
With tracing disabled
Screenshot_20250801_133130

---
Without this branch
Screenshot_20250801_133225

idursun:
@vic That's because your "matched" style has underline set and lanes are drawn using "revisions matched" style, so it is inheriting the _underline_ attribute from it, that's why it renders like this. 

To get rid of this, you need to specifically set "revisions matched" with `underline = false`. That would be my recommendation, but unfortunately, that won't work either. I need to change the way the default attributes are inherited when the final style is calculated. Currently, overrides with default values have no effect (i.e. underline = false is the default value). I will create a separate PR to make this a little bit more predictable.

Meanwhile, you can remove the underline from the "matched" style and set it only where you want it to be underlined.

In this branch, I can set the underline to false explicitly as users will almost never want to have it on in the lanes part of the graph.

zerowidth:
Wow, I love this!

I noticed that only the graph lines and the "main" text gets dimmed, but not the rest of the change (change id, author, date) does not. Nor does any non-"plain" text, e.g. the "empty" or "no description" labels.

For debugging, I set `dimmed = "bright red"` to show which parts of the text are targeted by the "dimmed" rule:

image
image

Is it  or could it be possible to dim the rest of the change log entry?

idursun:
@zerowidth thanks!

Which terminal are you using? It's setting the `faded = true` but now I am thinking some terminals might not be supporting it. I am on ghostty. I will have a look at with the other ones, and come back to you.

zerowidth:
I'm using `iterm2`, and that led me to this setting:

image

I had "faint" set to 100% opaque. Reducing that value below 100% worked, now I see the inactive branches dimmed as expected. Thanks, and TIL this "faint" ANSI escape code!

vic:
@idursun thank you, I've [removed underline](https://github.com/vic/tinted-jjui/commit/fc2537d6b451cc07861c386ab564a10e0810fabb) and it works as expected.

I only see an issue now when the lane is forked and I'm hovering a child, the fork point is partially highlighted.

image


Also maybe this is something minor (and I could live with this), if I have "unrelated" branches, see my revset here, and I hover on the last one, the first one's lane is also highlighted. 

image


Thanks for your amazing work. This feature alone is beautiful ! <3

penryu:
Truly a beautiful feature! (And it's working great on my 1-line-per-rev template as well!)

(Shown after patching the theme with `underline = false` fix described above.)

image

vic:
Hey @idursun, could you help me with some insight as to which style should I change in `tinted-jjui` themes so the following looks nice?

My terminal has a solarized-light background, but the jjui theme has a light gray theme. 

image

idursun:
@vic I pushed a change now. It should use the correct styles (at least according to my testing)

vic:
nice! works great now. thank you!

idursun:
Since this is behind a configuration option and seems to be working fine in general, I will mark this PR ready for review and probably merge. Later on, I will focus on optimising it.

PR #252, opened 2025-08-01
feat: make log format and revset configurable

### LINKED ISSUE #217: FR: allow overriding the default revset log via configuration
In jj, I have the log show the default display of only a few changes. In jjui, since I have the whole screen dedicated to the program, it's a bit of a waste to show only two lines of log, however jjui doesn't let me configure the number of log items separately. It would be great if this would be a config item, so I could specify @all.

If this is indeed possible, maybe the docs can spell it out a bit more? I imagine it would be a popular request.

penryu:
One of the few times something that _sounded_ easy _was_ easy.

Not certain if `preview` is the right table to add the `log_command` field, but happy to move it if there's a better place for it.

Tested that it works with both 2-line and 1-line log formats. (My preferred is a 1-line, hence this PR.)

Resolves #217

### DIFF
--- a/internal/config/config.go
+++ b/internal/config/config.go
@@ -104,6 +104,7 @@ type UIConfig struct {
 
 type RevisionConfig struct {
 	LogCommand []string `toml:"log_command"`
+	Revset     string   `toml:"revset"`
 }
 
 type PreviewConfig struct {
--- a/internal/config/config.go
+++ b/internal/config/config.go
@@ -17,6 +17,7 @@ var Current = loadDefaultConfig()
 type Config struct {
 	Keys                           KeyMappings[keys] `toml:"keys"`
 	UI                             UIConfig          `toml:"ui"`
+	Revision                       RevisionConfig    `toml:"revision"`
 	Preview                        PreviewConfig     `toml:"preview"`
 	OpLog                          OpLogConfig       `toml:"oplog"`
 	Graph                          GraphConfig       `toml:"graph"`
@@ -101,8 +102,11 @@ type UIConfig struct {
 	AutoRefreshInterval int `toml:"auto_refresh_interval"`
 }
 
+type RevisionConfig struct {
+	LogCommand []string `toml:"log_command"`
+}
+
 type PreviewConfig struct {
-	LogCommand               []string `toml:"log_command"`
 	RevisionCommand          []string `toml:"revision_command"`
 	OplogCommand             []string `toml:"oplog_command"`
 	FileCommand              []string `toml:"file_command"`
--- a/internal/config/config.go
+++ b/internal/config/config.go
@@ -17,7 +17,7 @@ var Current = loadDefaultConfig()
 type Config struct {
 	Keys                           KeyMappings[keys] `toml:"keys"`
 	UI                             UIConfig          `toml:"ui"`
-	Revision                       RevisionConfig    `toml:"revision"`
+	Log                            LogConfig         `toml:"log"`
 	Preview                        PreviewConfig     `toml:"preview"`
 	OpLog                          OpLogConfig       `toml:"oplog"`
 	Graph                          GraphConfig       `toml:"graph"`
@@ -102,9 +102,9 @@ type UIConfig struct {
 	AutoRefreshInterval int `toml:"auto_refresh_interval"`
 }
 
-type RevisionConfig struct {
-	LogCommand []string `toml:"log_command"`
-	Revset     string   `toml:"revset"`
+type LogConfig struct {
+	Template string `toml:"template"`
+	Revset   string `toml:"revset"`
 }
 
 type PreviewConfig struct {
--- a/internal/config/config.go
+++ b/internal/config/config.go
@@ -17,7 +17,7 @@ var Current = loadDefaultConfig()
 type Config struct {
 	Keys                           KeyMappings[keys] `toml:"keys"`
 	UI                             UIConfig          `toml:"ui"`
-	Log                            LogConfig         `toml:"log"`
+	Revisions                      RevisionsConfig   `toml:"revisions"`
 	Preview                        PreviewConfig     `toml:"preview"`
 	OpLog                          OpLogConfig       `toml:"oplog"`
 	Graph                          GraphConfig       `toml:"graph"`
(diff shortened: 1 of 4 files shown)


penryu:
Interestingly, this seems to require `--color always`; trying with `--color never` results in an empty revset (presumably covering an underlying failure).

I do want color, but it seemed interesting enough to mention.

vic:
I believe coloring is required because that's how jjui extracts commit_ids (I believe our screen parser does some ansi color magic trying to find stuff) not sure that's the cause, tho. If removing coloring breaks the revisions ui, shall we always add `--color always` to the command array that can be changed by the user?



> Not certain if `preview` is the right table to add the `log_command` field, but happy to move it if there's a better place for it.

Commands in that table are used to render the preview pane (on different types of content). So in theory there must be a better place since strictly speaking this change affects not the preview but the revisions view. But I guess we can move with it unless idursun thinks differently.

penryu:
> Commands in that table are used to render the preview pane (on different types of content). So in theory there must be a better place since strictly speaking this change affects not the preview but the revisions view. But I guess we can move with it unless idursun thinks differently.

Yeah, I was thinking something similar. I just wasn't sure if it was worth spinning up a new table, or just stash it in the top-level. Ultimately I used `preview` just because it had a lot of similar fields already.

I'll make a separate revision table in a second commit and we can abandon or squash as people prefer.

penryu:
"Revision" feels a little ambiguous here, especially as there's a `RevisionCommand` in the `PreviewConfig`, so I won't complain if people want to rename things. 🙃

idursun:
Hey @penryu thanks for this!

Is this PR addressing #217?

penryu:
> Hey @penryu thanks for this!
> 
> Is this PR addressing #217?

Ah, no, this is just to scratch my own itch to get a denser log output. It doesn't address the revset used at all... yet.

It wouldn't be much trouble to add a `revset` key to the `revision` table already implemented.

The only thing I'm looking for now is where to load it...

penryu:
Apologies. I attempted to resolve some conflicts in the web UI I accidentally closed this.

penryu:
Latest push adds `revset` to the new `revision` TOML table. The priority is as follows:

1. value passed with `-r` flag at command line
2. value of `revision.revset` in jjui/config.toml
3. value of `revsets.log` in jj/config.toml

This allows the user to specify a different default revset for jjui, or fall back to the default from jj, and so should address #217

I'm a _little_ concerned that the `revision` table now _has_ a `revset` field, but it's not provided in the default toml. Will that cause confusion, or is the comment I added sufficient? Is there a better way to handle this?

internal/config/default/config.toml
```diff
@@ -110,6 +110,10 @@ limit = 0
 
 [ui.colors]
 
+[revision]
+  log_command = ["log", "--color", "always", "--quiet"]
```
idursun: Can we get rid of the log command customisation bits because it is central to `jjui` and shouldn't be customisable.

internal/jj/commands.go
```diff
@@ -32,7 +34,7 @@ func ConfigListAll() CommandArgs {
 }
 
 func Log(revset string, limit int) CommandArgs {
-	args := []string{"log", "--color", "always", "--quiet"}
+	args := config.Current.Revision.LogCommand
```
idursun: ditto

penryu: The original string slice has been restored.

internal/config/default/config.toml
penryu: Ah, that is a good point. In the end, all I really wanted was to change the template used to render the log — jj's `templates.log` key by default.

In light of this, I removed the `log_command` key and refactored a bit:

- The table is renamed to `log`, as both keys it contains deal with arguments to `jj log`. (An alternative might also be `revlog`?)
- The original string slice of arguments is restored and doesn't change.
- `[]string log_command` becomes `string template` and, if set, `["-T", Log.Template]` is appended to args.

This should have less risk of breaking log parsing (`--color always --quiet` are still included), and while it may still be possible for `jjui.log.template` to be missing data that jjui needs to parse out, this is no less possible when users set `templates.log`. And it allows the user to set a jjui-specific value, just like in #217.

Does this seem reasonable as implemented?

(Full disclosure: the whole reason I opened this PR is that I have a log template that renders one-change-per-line, and I would _love_ to be able to use in jjui.)

penryu: (Of course, I realize I could have just set `jj.templates.log` and done this anyway. However, I'm already thinking of things I would like to tweak about it to make it even better in jjui.)

idursun: Could you share your log template with me? I’d love to see how it looks in jjui and see if there’s a simple thing we can change to make it work.

internal/config/config.go
```diff
@@ -17,6 +17,7 @@ var Current = loadDefaultConfig()
 type Config struct {
 	Keys                           KeyMappings[keys] `toml:"keys"`
 	UI                             UIConfig          `toml:"ui"`
+	Log                            LogConfig         `toml:"log"`
```
idursun: I'd personally change this to `Revisions` because we refer that part of that app as revisions.

internal/config/default/config.toml
penryu: Ah certainly.

```toml
onelog_commit_first = '''
if(root,
  format_root_commit(self),
  label(if(current_working_copy, "working_copy"),
    concat(
      separate(" ",
        format_short_change_id_with_hidden_and_divergent_info(self),
        if(author.email(), author.email().local(), email_placeholder),
        format_timestamp(commit_timestamp(self)),
        format_short_commit_id(commit_id),
        bookmarks,
        tags,
        working_copies,
        if(git_head, label("git_head", "git_head()")),
        if(conflict, label("conflict", "conflict")),
        if(empty, label("empty", "(empty)")),
        if(description,
          description.first_line(),
          label(if(empty, "empty"), description_placeholder),
        ),
      ) ++ "\n",
    ),
  )
)
'''
```

So at the moment my `jj/config.toml` has `templates.log = 'onelog_relative'`, and it is working in jjui as well.

All I really wanted to accomplish with this PR was to let me use a slightly different template in jjui.

internal/config/config.go
penryu: Ah, that's great feedback! I'll change it back.

PR #254, opened 2025-08-01
feat(fuzzy_input): More intuitive movement on history and different suggestions modes.

vic:
This patch makes the following changes to the input for `jj exec`/`sh exec`: 

- When entering exec mode, fuzzy suggestions are OFF. This is more intuitive as you would expect on entering a system shell's prompt. Where you can use `ctrl+p`, `ctrl+n` to move on immediate history instead of moving on suggestions. 
- Suggestions are enabled with `ctrl+r`, again a common idiom on shells. First suggestion mode is fuzzy since it will most likely give the expected result.
- Using `ctrl+r` again switches to regex match, this is useful when trying to match more continuous strings than the fuzzy match algorithm.
- Using `ctrl+r` starts the cycle again by turning suggestions off.

Curently `ctrl+r` is not expected to be a customized keybinding (just like we can't directly customize the textinput keybindings for moving on prev/next suggestions).

### DIFF
--- a/internal/ui/fuzzy_input/fuzzy_input.go
+++ b/internal/ui/fuzzy_input/fuzzy_input.go
@@ -2,9 +2,11 @@ package fuzzy_input
 
 import (
 	"fmt"
+	"regexp"
 	"strconv"
 	"strings"
 
+	"github.com/charmbracelet/bubbles/help"
 	"github.com/charmbracelet/bubbles/key"
 	"github.com/charmbracelet/bubbles/textinput"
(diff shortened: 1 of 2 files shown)


idursun:
Hey, thanks for this.

> When entering exec mode, fuzzy suggestions are OFF. 

When they are turned off, `j`/`k` move up and down in the history, effectively clearing what I typed already. Seems like a bug to me.

vic:
@idursun fixed now :) It was a bug indeed.

idursun (approved):
Thanks! LGTM.

PR #255, opened 2025-08-03
feat(force_apply): keybinds for using --ignore-immutabe.

### LINKED ISSUE #248: FR: Modifier to allow `--ignore-immutable` on operations.
**Is your feature request related to a problem? Please describe.**

Some operations on immutable revisions are denied by default by jj. This is a good thing, however sometimes we want to use `--ignore-immutable` flag as we do on jj cli. 

**Describe the solution you'd like**

Most actions like Rebase, Abandon, etc have a key-binding to confirm the operation. My proposal is to use `alt+` modifier to make it "force" (ie, use --ignore-immutable). So for example on the dialog for abandon I'd use `alt+enter` to confirm. 

**Describe alternatives you've considered**

Currently whenever I need to ignore immutable commits I simply use `:` jj exec.

**Why current features fail to cover this request**

It is possible using jj exec, but I resort to jj exec only once I get an error by using ui workflows.

**Additional context**
N/A.

vic:
Currently applies to:

- Abandon
  Using alt+enter on confirmation popup

- Rebase commit
  Using alt+enter

- Edit commit
  Using alt+e on a revision.

Fixes #248


internal/config/default/config.toml
```diff
@@ -7,7 +7,7 @@ limit = 0
   jump_to_parent = ["J"]
   jump_to_children = ["K"]
   jump_to_working_copy = ["@"]
-  apply = ["enter"]
+  apply = ["enter", "alt+enter"]
```
idursun: I think this should be a separate key binding. The reason for that is that we want user to press it willingly. I know later on in the code you are checking if the `Alt` modifier is active but then if the user changes this key bindings to something else that doesn't have `alt` in it that it won't work.

internal/config/default/config.toml
```diff
@@ -18,7 +18,7 @@ limit = 0
   quit = ["q"]
   help = ["?"]
   describe = ["D"]
-  edit = ["e"]
+  edit = ["e", "alt+e"]
```
idursun: Similarly here, `alt+e` is not captured on my Mac and I would have to change this to something else, then because of the `Alt` check, it wouldn't work.

idursun (changes requested):
Thanks!

I had added some comments about the usage of the `apply` keybinding. I think it should be a separate keybinding named something like `force_apply`.

vic:
> I had added some comments about the usage of the `apply` keybinding. I think it should be a separate keybinding named something like `force_apply`.

right! having separate key-bindings makes sense. I'll update PR with this changes.

internal/ui/confirmation/confirmation.go
```diff
@@ -51,7 +52,13 @@ func WithStylePrefix(prefix string) Option {
 // WithOption adds an option to the confirmation dialog
 func WithOption(label string, cmd tea.Cmd, keyBinding key.Binding) Option {
 	return func(m *Model) {
-		m.options = append(m.options, option{label, cmd, keyBinding})
+		m.options = append(m.options, option{label, cmd, keyBinding, cmd})
+	}
+}
+
+func WithAltOption(label string, cmd tea.Cmd, altCmd tea.Cmd, keyBinding key.Binding) Option {
```
idursun, line 59: This is fine for now, but I will probably revisit this because it is not configurable and also doesn't work on my Mac, which is fine as it is not taking anything away functionality wise.

internal/config/keys.go
```diff
@@ -14,6 +14,7 @@ func Convert(m KeyMappings[keys]) KeyMappings[key.Binding] {
 		JumpToChildren:    key.NewBinding(key.WithKeys(m.JumpToChildren...), key.WithHelp(JoinKeys(m.JumpToChildren), "jump to children")),
 		JumpToWorkingCopy: key.NewBinding(key.WithKeys(m.JumpToWorkingCopy...), key.WithHelp(JoinKeys(m.JumpToWorkingCopy), "jump to working copy")),
 		Apply:             key.NewBinding(key.WithKeys(m.Apply...), key.WithHelp(JoinKeys(m.Apply), "apply")),
+		ForceApply:        key.NewBinding(key.WithKeys(m.ForceApply...), key.WithHelp(JoinKeys(m.ForceApply), "force apply")),
```
idursun, line 17: New keys needed to be mapped here too.

idursun (approved):
This is good to go now. See the comments.

I will merge this to prevent conflicts that are happening on the other PR that's also touching confirmation dialog code. #256

internal/config/keys.go
vic: thanks, somehow forgot that!

azzamsa:
> Currently applies to:
> 
>     * Abandon
>       Using alt+enter on confirmation popup
> 
>     * Rebase commit
>       Using alt+enter
> 
>     * Edit commit
>       Using alt+e on a revision.
> 
> 
> Fixes #248

Any reason why squash is not included?

PR #256, opened 2025-08-03
feat(confirmation): Show key-bindings on confirmation dialog.

vic:
This helps the user to not having to move using direction keys, but directly use shortcuts instead.

Buttons are moved to their own line since now they occupy more space due to them including key-bindings help.

### DIFF
--- a/internal/ui/confirmation/confirmation.go
+++ b/internal/ui/confirmation/confirmation.go
@@ -39,6 +39,32 @@ type Model struct {
 	stylePrefix string
 }
 
+func (m *Model) ShortHelp() []key.Binding {
+	var bindings []key.Binding
+	for _, option := range m.options {
+		bindings = append(bindings, option.keyBinding)
+	}
+	bindings = append(bindings,
+		key.NewBinding(
+			key.WithKeys("left", "h"),
+			key.WithHelp("←", "left"),
+		),
+		key.NewBinding(
+			key.WithKeys("right", "l"),
+			key.WithHelp("→", "right"),
+		),
+		key.NewBinding(
+			key.WithKeys("enter"),
+			key.WithHelp("↵", "select"),
+		),
+	)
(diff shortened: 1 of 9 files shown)


idursun (changes requested):
Hey @vic 

Thanks for working on this.

I appreciate the effort here. I had a look and this is how it renders for me:

image

Shortcut is not very eligible as they selected foreground is very close to the shortcut key foreground. To make it eligible, we would need to change the default theme, which I don't want to.

Another concern is that the options have become very wordy and long, and that's why you had to move those down. I like the new layout though.

Instead may I suggest that these shortcuts to these menu items are shown in the status like every other key bindings we show. 

Currently, abandon doesn't show anything but it can easily show the `y` and `n` bindings there. The same approach would work for details and with some effort for undo too.

Any model/operation that's implementing `help.KeyMap` can give a list of shortcuts to be displayed in the status bar. Even better, it can be implemented on the `confirmation` model so that the underlying models can query if they are showing the confirmation, and if so, they return the list of the confirmation, and if not they return their own list. So, everything can change dynamically.

image

I've actually did that work and seems to be working fine and it is aligned with the rest of the UI. I can push those changes to this branch and remove the shortcut rendering in the confirmation options.

This would be my suggestion if I am not missing anything.

vic:
> Instead may I suggest that these shortcuts to these menu items are shown in the status like every other key bindings we show.

makes a lot of sense to me!


> I can push those changes to this branch and remove the shortcut rendering in the confirmation options.

Please push :)

vic:
> This would be my suggestion if I am not missing anything.

Only thing I might suggest when showing shortcuts on the status bar, perhaps we should show movement keys last -- having the keybindings for the confirmation actions first. e.g.: `y` and `n` first and direction keys, enter, alt+enter last.

PR #257, opened 2025-08-03
feat(set_bookmark): Accept with `ctrl+s` just like inline_desc.

vic:
This patch makes the set_bookmark operation use the same keybindings as inline_desc to accept input.

This improves consistency as user would have expected `ctrl+s` to work on inline edit for bookmarks as well.

### DIFF
--- a/internal/ui/operations/bookmark/set_bookmark.go
+++ b/internal/ui/operations/bookmark/set_bookmark.go
@@ -1,10 +1,13 @@
 package bookmark
 
 import (
-	"github.com/charmbracelet/bubbles/textinput"
 	"strings"
 
+	"github.com/charmbracelet/bubbles/key"
+	"github.com/charmbracelet/bubbles/textinput"
+
 	tea "github.com/charmbracelet/bubbletea"
+	"github.com/idursun/jjui/internal/config"
 	"github.com/idursun/jjui/internal/jj"
 	"github.com/idursun/jjui/internal/ui/common"
 	"github.com/idursun/jjui/internal/ui/context"
@@ -41,12 +44,13 @@ func (s *SetBookmarkOperation) IsFocused() bool {
 }
 
 func (s *SetBookmarkOperation) Update(msg tea.Msg) (operations.OperationWithOverlay, tea.Cmd) {
+	km := config.Current.GetKeyMap()
 	switch msg := msg.(type) {
 	case tea.KeyMsg:
-		switch msg.String() {
-		case "esc":
+		switch {
+		case key.Matches(msg, km.Cancel):
 			return s, common.Close
-		case "enter":
+		case key.Matches(msg, km.Apply, km.InlineDescribe.Accept):
 			return s, s.context.RunCommand(jj.BookmarkSet(s.revision, s.name.Value()), common.Close, common.Refresh)
 		}
 	}


idursun:
This is already possible by adding another key to the `apply` keybinding.

```toml
apply = ["enter", "ctrl+s"]
```

This should make it available across the board. The only reason that inline describe has an additional key binding is because `enter` is bound to "new line", hence cannot be used.

It also looks a little bit weird that set bookmark is using a keybinding from inline describe.

vic:
I was not sure about adding `ctrl+s` to `accept`, because perhaps `accept` is used elsewhere (like confirmation modals) or other places that might use `accept` (in my mind that is `enter`) keybinding. `ctrl+s` is more like "save" and is mentally ok for inline editing bookmarks and descriptions, but not for other places that might be expecting `accept` as `enter`, like confirmations / jj shell / other places we might be using `enter`.

That's why I did not go this route. And instead preferred to re-use whatever key-binding the user already has for inline_desc. Does this makes sense?

idursun:
> I was not sure about adding `ctrl+s` to `accept`, because perhaps `accept` is used elsewhere (like confirmation modals) or other places that might use `accept` (in my mind that is `enter`) keybinding. `ctrl+s` is more like "save" and is mentally ok for inline editing bookmarks and descriptions, but not for other places that might be expecting `accept` as `enter`, like confirmations / jj shell / other places we might be using `enter`.

Yes, that's correct and fair. While editing text, `enter` should be the default for accepting changes across the board. We don't even need to make it configurable. 

The only inconsistent case is the inline describe editor, which is a multi-line editor. If set bookmark editor was a multi-line editor, then this would make sense. I've added added extra `ctrl+s` binding there just in case users have a problem with using `alt+enter` due to limited terminal key support. If that concern is unfounded, then the move to make for consistency sake is to remove the `ctrl+s` extra binding from inline describe.

What do you think?

vic:
Having `ctrl+s` on multi-line editors (inline_desc) makes total sense.  I just got used to it actually, and was surprised when trying to add a bookmark that `ctrl+s` was doing nothing. But it was just because somehow my mind expected both to work the same way (again as `save`). 

I know bookmarks are not multi-line, perhaps all we are missing is to show the keybinding help at the status bar when entering bookmark edit mode. Without any key indicator my mind just assumed that `ctrl+s` was going to work as it did on inline_desc.

I'd say to remove `alt+enter` instead and just leave `ctrl+s` on multi-line editors. Since `alt+` seems limited on Mac terminals ? (I'm on Linux so I haven't tested this first-hand).


So, what if we use `ctrl+s` for saving inline-editors and also show help on the status bar. 

(we might remove alt+enter from inline_desc editor)

(on bookmark_edit that is not multi-line should `enter` be still bound?)

idursun:
> I'd say to remove alt+enter instead and just leave ctrl+s on multi-line editors. Since alt+ seems limited on Mac terminals ? (I'm on Linux so I haven't tested this first-hand).

`alt+enter` works fine on Macs. Only `alt+letter` bindings don't work, at least with my current keyboard settings.

> (on bookmark_edit that is not multi-line should enter be still bound?)

Yes, because `enter` doesn't mean new line in single line editors, so it should stay as the default for finishing editing.

> Having ctrl+s on multi-line editors (inline_desc) makes total sense. I just got used to it actually, and was surprised when trying to add a bookmark that ctrl+s was doing nothing. But it was just because somehow my mind expected both to work the same way (again as save).

Going back to my first comment, to support this, we should respect the following configuration in single-line text editors, **not only in set bookmark but in all of them (e.g. revset, menu filters, status editors, etc)**

```toml
apply = ["enter", "ctrl+s"]
```

To reiterate, instead of patching one use case after another, I'd personally go with hunting down every inlined `enter` key binding, and change it with `apply`.

vic:
Closing this, I will first make sure we all places that use an inlined `enter` use the `apply` keybinding instead.

PR #267, opened 2025-08-13
fix(leader): Dont split sent keys to prevent percieved slowness.

### LINKED ISSUE #266: Leader key send is very slow

1. Use a leader key binding with send and e.g. `[":" "rebase -r $change_id"]`

## Expected Behavior

Command gets executed immediately.

## Actual Behavior

Words like "$change_id" slowly appear letter by letter.

## Specifications

- Platform: nixos
- Version: 0.9.1

vic:
Fixes #266

### DIFF
--- a/internal/ui/leader/leader_test.go
+++ b/internal/ui/leader/leader_test.go
@@ -29,7 +29,7 @@ func Test_sendCmds_detects_tea_keytypes(t *testing.T) {
 }
 
 func Test_sendCmds_produces_rune_keys_on_non_keytypes(t *testing.T) {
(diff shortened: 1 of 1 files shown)


vic:
> [!Warning]
> This patch breaks the following:  

```toml
# this used to work before this change
# open git modal, fetch, / (search), type "all"
send = ["gf/all"]
```

Using this patch, this leader key would need to be rewritten for it to work, like:

```toml
send = ["g", "f", "/", "all"]
```

This is because sending "gf/all" as a single Key (runes) command does not trigger the git menu, fetch, etc...

PR #268, opened 2025-08-13
fix(leader): Use tea.Sequence instead of our custom cmd sequencing.

vic:
Related: #266


maralorn:
You people are awesome!

PR #283, opened 2025-08-27
feat: squash files from details view

Jollywatt:
This might not be worth adding, since the existing squash command is more general — so you may take it as a suggestion.

I often want to squash file(s) from the current revision into the parent. When using JJUI, I usually enter the details view first to see what's going on, then I select the files I want to squash. However, there is no squash operation in details mode: you must exit details, select the source revision, press `S`, select the parent, then interactively select the files.

This PR adds a shortcut `S` to the details view that squashes the selected files into the parent directly.

Limitations:
- You can only select at the level of files, not lines
- Squashing is always into the parent (if the parent is unique)

Despite its narrow scope, I use this command all the time, so perhaps it is worth sharing.

### DIFF
--- a/internal/config/default/config.toml
+++ b/internal/config/default/config.toml
@@ -62,6 +62,7 @@ limit = 0
     mode = ["l"]
     close = ["h"]
     split = ["s"]
+    squash = ["S"]
     restore = ["r"]
     absorb = ["A"]
     diff = ["d"]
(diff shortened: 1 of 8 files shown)


nickchomey:
If I understand this correctly, I think it is useful but would be much better if you had to select which revision to squash into. Rather than automatically into the parent.

Jollywatt:
> would be much better if you had to select which revision to squash into

I agree. I'm not familiar enough to know how to implement that with `jjui`, since all the other commands seem to require selecting revisions as the first step, before invoking them.

nickchomey:
Hopefully someone else can carry the torch forward!

idursun:
Hey, thanks for this. 

I will look at implementing choosing a target to squash into.

idursun:
@Jollywatt I pushed a commit that allows you to choose the target revision to squash into. Let me know what you think. We can merge this if you are OK with it.

Jollywatt:
That works amazingly! It's so cool to be able to just "pick up" files and drop them into other revisions just like that!

I guess in the future we could add an interactive mode which you toggle with `i`, just like the other squash command. Then the only difference between the two squashes would be the order in which you select things. I think it's good to have both: sometimes you are thinking files-first and other times revisions-first. It's very cool to support both so nicely.

PR #297, opened 2025-09-04
feat(describe): Allow multiple revisions

the-kenny:
`jj describe` takes multiple revsets to describe. This change allows users to describe multiple revisions simultaneously by first selecting them via `space`, followed by `D`.


idursun (approved):
This is cool! Thank you very much for contributing.

PR #303, opened 2025-09-15
feat(open_file): open one or more files from 'detail' view

the-kenny:


idursun:
Hey, thanks for this.

Even though there's not a description or agreed upon issue, let me give my feedback.

This is named open files but it is actually opening an editor. This comes with the baked in assumption that the user can actually edit the file in any revision, which is not the case. This will only edit the file (if it exists at all) in the working copy, but not the in the selected revision. 

I can see this being useful if it was reframed as "view file" and opened the file viewer (diff viewer can be reused for this) to display the contents of the file in that revision. Sometimes I find myself needing to see the whole file rather than the diff of the file.

the-kenny:
Your criticism is right - this was a one-off change I did a few days ago which I haven't spent very much time into planning.

I originally built this because I was in need for a shortcut to jump to the selected file in my editor (I'm running `jjui` inside Zed). I didn't consider how the feature would work in a different setup.

Another idea that crossed my mind offering an entry point for `fuzzy file search`, pre-populating the input field with the path to the selected file. 

Feel free to close this - and if you're open for another approach I'll start a discussion first.

vic:
Related #219. 

@the-kenny you can achieve `edit file` using a [leader key](https://github.com/idursun/jjui/wiki/Leader-Key#edit-a-file-from-revision-detail-idea-from-184). The example there uses `$file` for the file under cursor, but you can easily use `$checked_files[@]` instead if you want to open multiple files marked with space.

the-kenny:
@vic Thanks, that's a great solution! Shame on me for not reading the docs about leader keys first.

Issue #109, opened 2025-06-08
FR: an option and binding to toggle UI


ilyagr:
I'd like an option and a key binding to toggle showing the "extra" lines of the UI (the top revset line and the bottom status/help line). This would improve the experience of using `jjui` instead of https://github.com/blacknon/hwatch to visualize my jj log in a separate `tmux` pane.

Some possible bindings would be `~` or backtick or `t`. The option could be `ui.show_ui_on_startup`. It'd also be nice if there was a command-line option (or a way to set arbitrary options from the command-line).

There's also a question of whether it should be possible to hide the status line without hiding the revset line. I'd lean towards "no" for now, for simplicity.

idursun:
Sorry I am having hard time understanding the reason for this feature. Are those extra lines causing distraction for you?

ilyagr:
They take up space, which is a bit limited if jjui shares the screen with the terminal where I am working.

It looks something like this.

<img width="1192" alt="Image" src="https://github.com/user-attachments/assets/72697bc7-4101-4284-83c4-67b786d40f3f" />

But I'm not insisting on it if it doesn't sound compelling. At worst, I might just make a patch for myself at some point and see how that goes.

shoce:
i double this request. sometimes would be really handy to see only changes list without that static information on the top and bottom rows. actually would like to have a config option to hide the bottom line, i already learnt the shortcuts i need.

Issue #68, opened 2025-05-01
Feature: Show the filepath (or name of cwd) of the current jj repo at the top


nickchomey:
Sometimes I have multiple jjui instance open in different panes of a full-screen terminal. There's no indication of which repo is which, so I keep having to look at the file diffs and messages to identify them.

Would it be possible to show at the top either the full filepath or at least the name of the current working directory?

![Image](https://github.com/user-attachments/assets/17420a64-3f51-44cf-a87d-9f5546c3dd6f)

idursun:
Hey, thanks!

The easiest way I can add that information is to the window title like so:

<img width="606" alt="Image" src="https://github.com/user-attachments/assets/66fdecad-8356-49aa-a97b-bf9ff0fd46f1" />

I am not sure if it would work for you too depending on what kind of terminal you are using (some don't show title, some do)

nickchomey:
I'm not sure if that would work or not. I use windows terminal and vs code terminal. I can test it if you push a commit for it

What I had in mind was putting the path above the revset line at the top, but it is all up to you, of course.

idursun:
Sure, it's in `main` now.

nickchomey:
Thanks! I just tried it - unfortunately it doesnt work for me. 

In Windows Terminal, it only shows the title in the main tab. But the panes themselves don't have tab headers, so you have to click on each pane in order to change the title shown in the tab. But the title gets significantly truncated by the tab size, so it doesnt matter anyway. I suppose you could show just the present directory rather than the full path, but it would only be a partial solution since it doesnt show per pane. 

![Image](https://github.com/user-attachments/assets/ed79b133-abbf-411f-8d7a-96779875399b)

The new jjui title mechanism doesnt show in the VS Code terminal, but it doesnt need to as it already shows the path from which it was run 

![Image](https://github.com/user-attachments/assets/a75c2d6f-227c-4d27-b518-654dc13f4cd3)

If you ever become inclined to add the cwd to the actual JJUI output, that would be great!

idursun:
> If you ever become inclined to add the cwd to the actual JJUI output, that would be great!

I will have a look at this but then you will only know what repository you are working on only after you performed a jj action.

Another option is to add it to the top and after the revset, something like 

`revset: present(@) | ancestors(immutable_heads().., 2) | present(trunk()) in ~/respositories/jjui`

nickchomey:
[...]

Issue #58, opened 2025-04-21
Feature: Autocommit feature


nickchomey:
I just had a situation where I was working away for a while and I wanted to rollback some of the effort. However, I forgot to create new change or even use jj at all for a couple hours. So, the diff in my current change was rather large. 

The evolog wasn't of use, because jj only creates a new commit (not change/revset) each time you run any jj command.

So, I started thinking that perhaps jjui (which I had open in a terminal that I wasn't using) could make auto commits by just running something like `jj log` (or whatever is already used to refresh) on some sort of configurable interval (eg 1 min). If i had had that, I would have been able to roll back to any point quite granularly. 

An alternative would be to use a filesystem watcher - watch the repo and make a commit any time a file changes. That seems like it would be excessive though.

idursun:
Hey, thanks for the feedback.

You have already linked to the previously discussed idea. I don't think creating a commit here is the best option, but I can see where you are coming from. `jj` creates a snapshot of the working copy (a.k.a an evolog entry) when interacted, and `jjui` is only reflects this when manually refreshed. 

I can add an option to auto refresh when `jjui` stays idle for X minutes. This can help with the problem you faced to some extent, but it's only a defensive workaround and not an ultimate fix.

nickchomey:
In my basic testing, running any jj command - eg `jj log` or even just using Ctrl+R to refresh jjui - will create a new commit ONLY IF something has actually changed in the repo. 

So, it seems to me that there would be no downside really to just running any jj command - it might as well be whatever jjui uses for the manual refresh - on a configurable interval. I'd probably try something like 10 seconds to start with, and slow it down if needed. But it could default to 0, so that it just never auto-refreshes unless you specifically set that config.

In short, just run the command that is currently triggered by Ctrl+R/manual refresh in an indefinite loop on some configurable interval... Seems like a fairly easy thing to implement? Or am I missing something about how jjui and/or jj work?

If something like this wont be able to be implemented in jjui, i suppose I could just set up some sort of cronjob or whatever to just run `jj log` on an interval.. But it would be really nice to have it in jjui, especially if it also solves the autorefresh issue #30.

nickchomey:
[...]

Issue #30, opened 2025-03-27
Automatically refresh the view


guibou:
When doing a `jj` operation outside of `jjui`, my `jjui` is not in sync anymore with the repository state. I would like `jjui` to auto refresh.

Could `jjui` register a file watcher (such as `inotify` on linux, but it exists multi platform) on the  `.jj` directory (I think it would be enough, maybe the watcher can even be more selective).

For example, on linux, I tried:

```
inotifywait --monitor -r .jj -e modify
```

(note that I do not suggest doing a blind recursive watch on the `.jj` directory, that's thousands of file and this could reach the inotify limit, but as a PoC, it shows some interesting results)

And any `jj` command, such as `jj edit` or `jj new` triggers an event and jjui could be refresh accordingly.

Thank you.

idursun:
Hey, thanks for the suggestion!

This is indeed a good idea but very tricky to get it right in a cross platform way. It should auto refresh only if changes are done outside of `jjui` and not when it is done via `jjui` as that would cause double refreshes.

Additionally, auto refresh may not even work because `jj` snapshots the working copy only when `jj log` or `jj status` commands are run (other than the obvious one like `jj describe`, `jj rebase`, etc). 

I will have a look into this to see what's possible and reliable.

nickchomey:
I wonder if jjui could simply refresh on a configurable interval, which would be much simpler than a filesystem watcher? At a glance, it only creates a new commit/snapshot if something has changed, so the interval could even be something fairly small (eg 5 seconds), and would avoid the issue with jjui causing double refreshes

This is related to the feature request (#58) that I just opened  

Though, if you do end up wanting to do a filesystem watcher, you might consider this watcher: https://github.com/e-dant/watcher. Frankenphp (a popular new golang webserver for php, built as a caddy module) [settled on it](https://github.com/dunglas/frankenphp/pull/1013#issuecomment-2336736615) after evaluating a lot of other options. The conversation there has a lot of info - I linked to a particularly useful comment.

nickchomey:
@guibou FYI, a new auto-refresh feature was introduced in the [latest commit to jjui](https://github.com/idursun/jjui/commit/6e27fe86137c122e7f5eb807f79a8856aee0977a). You can set autorefresh interval for jjui to refresh on. 

I've set it to 15 seconds for now. 

[...]

PR #306, opened 2025-09-18
feat: remove auto refresh in favour of focus tracking

idursun:
Auto-refresh, which relied on refreshing revisions on a fixed interval configured via `ui.auto_refresh_interval`, worked most of the time but was also the source of many bugs that annoyed me for a while. It’s also turned off by default, so many people, including myself, don’t even use it.

This PR removes auto-refresh in favour of focus tracking. Basically, whenever your focus moves away from your terminal and then comes back, `jjui` will refresh itself. This way, if you’ve made changes somewhere else, you’ll always be presented with the latest state of your repo. This approach is deterministic and much more reliable than auto-refresh, as it wouldn’t refresh if you happened to focus back between the refresh interval.

### DIFF
--- a/cmd/jjui/main.go
+++ b/cmd/jjui/main.go
@@ -38,7 +38,6 @@ func getVersion() string {
 
 var (
 	revset     string
-	period     int
 	limit      int
 	version    bool
 	editConfig bool
@@ -48,8 +47,6 @@ var (
 func init() {
 	flag.StringVar(&revset, "revset", "", "Set default revset")
 	flag.StringVar(&revset, "r", "", "Set default revset (same as --revset)")
-	flag.IntVar(&period, "period", -1, "Override auto-refresh interval (seconds, set to 0 to disable)")
-	flag.IntVar(&period, "p", -1, "Override auto-refresh interval (alias for --period)")
 	flag.IntVar(&limit, "limit", 0, "Number of revisions to show (default: 0)")
 	flag.IntVar(&limit, "n", 0, "Number of revisions to show (alias for --limit)")
 	flag.BoolVar(&version, "version", false, "Show version information")
@@ -180,9 +177,6 @@ func main() {
 	common.DefaultPalette.Update(appContext.JJConfig.GetApplicableColors())
 	common.DefaultPalette.Update(config.Current.UI.Colors)
 
-	if period >= 0 {
-		config.Current.UI.AutoRefreshInterval = period
-	}
 	if revset != "" {
 		appContext.DefaultRevset = revset
 	} else if config.Current.Revisions.Revset != "" {
@@ -192,7 +186,7 @@ func main() {
 	}
 	appContext.CurrentRevset = appContext.DefaultRevset
 
-	p := tea.NewProgram(ui.New(appContext), tea.WithAltScreen())
+	p := tea.NewProgram(ui.New(appContext), tea.WithAltScreen(), tea.WithReportFocus())
 	if _, err := p.Run(); err != nil {
 		fmt.Printf("Error running program: %v\n", err)
 		os.Exit(1)
--- a/internal/config/config.go
+++ b/internal/config/config.go
@@ -120,10 +120,7 @@ type TracerConfig struct {
 type UIConfig struct {
 	Theme  ThemeConfig      `toml:"theme"`
 	Colors map[string]Color `toml:"colors"`
-	// TODO(ilyagr): It might make sense to rename this to `auto_refresh_period` to match `--period` option
-	// once we have a mechanism to deprecate the old name softly.
-	AutoRefreshInterval int          `toml:"auto_refresh_interval"`
-	Tracer              TracerConfig `toml:"tracer"`
+	Tracer TracerConfig     `toml:"tracer"`
 }
 
 type RevisionsConfig struct {
--- a/internal/config/config_test.go
+++ b/internal/config/config_test.go
@@ -46,17 +46,6 @@ light = "light-theme"
 	assert.Equal(t, "light-theme", config.UI.Theme.Light)
 }
 
-func TestLoad_AutoRefreshInterval(t *testing.T) {
-	content := `
-[ui]
-auto_refresh_interval = 5000
-`
-	config := &Config{}
-	err := config.Load(content)
-	assert.NoError(t, err)
-	assert.Equal(t, 5000, config.UI.AutoRefreshInterval)
-}
-
 func TestLoad_Colors_StringAndObject(t *testing.T) {
 	content := `
 [ui.colors]
--- a/internal/config/default/config.toml
+++ b/internal/config/default/config.toml
@@ -109,7 +109,6 @@ limit = 0
 
 [ui]
   theme = ""
-  auto_refresh_interval = 0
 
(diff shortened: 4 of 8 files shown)


nickchomey:
Disappointing. I made heavy use of auto refresh by having jjui open in a separate monitor and I could see changes come in without having to focus on it.

idursun:
Sorry to hear that. Maybe it will come back in a different form.

I think the new way will benefit more users, especially the ones who never activated the auto refresh in the first place. Additionally, the focus tracking approach also solves some issues when the jjui was suspended and then resumed using `ctrl+z` as it will refresh itself automatically.

I will keep your use case in mind and bring it back in some form when the opportunity arises.

nickchomey:
Is there a technical reason why focus and optional auto-refresh can't co-exist? Leave auto refresh off by default and those of us who want it can use it.

idursun:
> Is there a technical reason why focus and optional auto-refresh can't co-exist?

Yes, it needs to be guarded against in many places if it happens to trigger while the user is interacting with `jjui`. Additionally, not much of a problem in small repos, but it is a UX headache when activated in big repos as the refresh takes a while. If a refresh is in progress after a command execution and then the auto refresh kicks in around that time then the first needs to be cancelled or conditioned so that the work is not duplicated, leading so many edge cases.

Secondly, it was solving the original ask for it at the wrong level. The original ask was for jjui to refresh automatically when they have done some changes to the repo outside of jjui. Hence, the auto refresh. But it didn't really solve the original ask  if they happen to move back to jjui between auto refresh cycle.

Thirdly, I didn't like the fact that you end up 20+ op log entries while you were coding away. Most of them are useless and polluting the op log history.

Lastly, to answer your question; it is straight forward to implement if it only works when jjui is out of focus (that means there will only be one message to be processed at any given time). Potentially, the UI can also change to add some useful information while in this mode and hide some of the parts UI (e.g. revset, and status) as they are only there to assist interacting with jjui. Opens up possibilities to implement #109 and your ask in #68.

nickchomey:
Fair enough! Thanks for your hard work on all of this

ilyagr:
Does this mean I'll have to click on `jjui` for the window to update?

This wouldn't be great for me. I often use `jjui` in a separate tmux tab, looking at how the commits are changing without ever selecting `jjui`.

image


I am also aware of some annoying concurrency jj bugs this triggers, but the tradeoff is worth it for me.

I'll probably keep using an old or patched version of jjui (well, I am already, but this might put an upper bound on how far I can upgrade). (But it's fine to keep this in jjui if you think it's best for your project, of course)

-----

It sounds like you also have some plans for improving this. I don't fully understand them, but they might be great as well. :)

nickchomey:
Yeah, I didnt really feel like elaborating, but the reasons provided for removing this were not all that convincing. 

* UX headaches - given that this was an opt-in feature, its my/our problem to deal with, not jjui's. 
* The original ask was something else - I was part of the discussion (#30, #58) of the original ask and explained this use case there. Plus, now that the use case was covered, there's less reason to remove it.
* 20+ op log entries - this is also my problem to deal with. Moreover, I use watchman to track ALL changes to ALL files. This is [officially documented](https://jj-vcs.github.io/jj/latest/config/#watchman), supported and even recommended by jj devs as it, apparently, improves performance on large repos. 
* issues with changes happening while refreshing - again, my problem. I think I run into this sometimes and just press Ctrl+R to refresh and carry on. 

The only actually valid reason - and it may very well justifiably supersede all of the others - is that it apparently causes a burden for the maintainers, and they don't want to carry it anymore. If so, understood! But I'd still like to request that the feature remain as-is, perhaps add a cautionary/"experimental" note somewhere, and then we simply deal with the issues that come with it.

vic:
friendly question: if you use jjui as a log viewer, without focusing it, but just having it on the side (or bottom - or another display), why not use a small shell loop that just renders the jj log continuously. Since you are using only for displaying changes as they are made from the jj cli and not interacting via jjui.

nickchomey:
I doubt anyone here is ONLY using it as a log viewer. We just leave it like that in the background until we do need to use it

ilyagr:
> friendly question: if you use jjui as a log viewer, without focusing it, but just having it on the side (or bottom - or another display), why not use a small shell loop that just renders the jj log continuously. Since you are using only for displaying changes as they are made from the jj cli and not interacting via jjui.

That's precisely what I did before I started using `jjui`. I'd use https://github.com/blacknon/hwatch (but see also a list of alternatives at https://github.com/blacknon/hwatch?tab=readme-ov-file#alternatives) and essentially did `hwatch jj log` (with some additional bells and whistles to show the current commit's `diff --summary`).

With jjui, I get that, but I can also switch into that window and (say) change a commit description very quickly (if it's a short description, <5 lines, say; for longer texts I prefer `jj describe`). Or, if both commits are visible, it's often faster to use `jjui` to rebase than to type it out in the CLI. Or I can use `jjui` to navigate the op-log and restore an old operation.

To be clear, while this use-case is important to me, I will absolutely be fine if auto-refresh is removed from jjui. I could move back to my previous workflow (and only open up `jjui` occasionally; it'd still be very useful), or I could keep using an old version of jjui.

I'm especially curious about what's possible if Ibrahim's intuition suggests that there's a better way. Take your time to cook!

ilyagr:
Aside for auto-refresh fans: in the lower-left corner of `jjui` in https://github.com/idursun/jjui/pull/306#issuecomment-3313990005, there's a spinner that looks like `⣯`. It's a feature I had an AI add to `jjui` to let me know whether the auto-updating is going OK, there's a little checkmark that appears next to it when `jjui` notices actual changes in the repo.

I never contributed it back because the AI code was/is ugly, and I kept meaning to clean it up before sending a PR. But it is a feature I quite like. If anyone wants to use it, see https://github.com/ilyagr/jjui/tree/spinner3. It's probably possible to rebase it onto the current main, at least it seems to rebase onto 19fc54a OK.

(I'm behind because I also use https://github.com/ilyagr/jjui/tree/mouse-zed2 which is harder to rebase, and I don't recommend it unless you're really interested. It has some bugs that can only be fixed in bubbletea).

ilyagr:
BTW, I think refresh on focus events is also a great idea. I'll use it. Perhaps it will substitute for auto-refresh for a good number of people.

Some caveats:
- Ideally, you'd want to disable focus event reporting when launching external commands (that might not recognize these sequences), and then (perhaps more importantly) enable focus events again when the external command is done. (I didn't check whether bubbletea or you are already doing that)
- tmux needs `set -g focus-events on` to enable focus events

idursun:
Thanks for the discussions folks!

I think I will revert this change for the time being to restore the existing functionality. 

Then I will make another change to add the focus tracking back. As far as I understand, auto refresh is mostly useful to folks who want to see the current state of the repo without interacting `jjui`. As a stretch goal, I will look into adding a "watch" view, where the interactive elements are hidden and but some visual indicators are added so that the user can see if and when auto refresh happens.

I believe this is a good middle ground:

* For those who don't have auto refresh enabled, focus tracking will help
* For those who have auto refresh enabled, then the existing functionality will there but be presented in a slightly better way to support that use case.

Let's see where this goes.

PR #317, opened 2025-09-26
feat: edit in editor escape hatch for inline describe editor

paradox460:
Adds an "edit in editor" binding that lets you bail out from the inline describe editor to a full fledged editor.

Inspired by similar functionality that I missed from LazyGit

https://github.com/user-attachments/assets/7b994fb0-b41c-48aa-81c7-7a36623cedb5


idursun (approved):
Thanks! Looks good to me

PR #318, opened 2025-09-29
feat: add copy description mode

paradox460:
Adds a "copy mode" to the revision list. Hitting a keystroke (default `ctrl+c`) brings up "copy mode" where you can copy the change id, commit id, description, or "full" describe message.

### DIFF
--- a/internal/config/default/config.toml
+++ b/internal/config/default/config.toml
@@ -107,6 +107,12 @@ limit = 0
     down = ["down"]
     accept = ["enter"]
     edit = ["alt+e"]
+  [keys.copy]
+    mode = ["ctrl+c"]
+    change_id = ["c"]
+    commit_id = ["i"]
+    description = ["d"]
+    full_info = ["f"]
 
 
 [ui]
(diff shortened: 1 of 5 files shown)


paradox460:
I thought about making the "full" description an editable template, but didn't. If that's something you're interested in, I can add it.

idursun:
@paradox460 Thanks working on this. Normally, we discuss new features in an issue and only implement it if the feature and the approach is agreed on.

I personally prefer `jjui` to only invoke `jj` as a command by default. Any command that's not `jj` is invoked either through custom commands or `exec_shell`. Both of them assume users implicit consent. 

I like the approach you take, and your implementation looks good. However, in the light of https://github.com/idursun/jjui/discussions/319 this can be implemented as a custom action.

paradox460:
No worries. Was just something that came up in discord discussion over the weekend. Doing it via scripting is a much more elegant approach!

idursun:
This is how I managed to do something similar in the scripting branch (syntax subject to change but it works).

https://github.com/idursun/jjui/blob/878fcbf31d8d587b92f3d24f3156bc0b7f2cd39c/internal/config/default/config.toml#L383

Issue #312, opened 2025-09-22
FR: Keybind to split a revision with the `--parallel` flag


ianjsikes:
## Is your feature request related to a problem? Please describe.
I'm working on a refactor that touches a ton of files in my codebase. I want to split these changes into many separate, parallel revisions, so my desired end-state would be something like this:
```
main -> a
    |-> b
    |-> c
    |-> d
    |-> e
    ...
```

The problem is that when you split in `jjui`, the new revision is placed as a parent to the current revision. So you end up with something like
```
main -> a -> b -> c -> d -> e -> ...
```
And then you have to individually rebase each commit back onto `main`.

## Describe the solution you'd like
The `jj` cli exposes a `--parallel` flag on the `split` command. I'd like a way to use this flag when splitting in `jjui`, so that the new revision and the current revision are not parent/child.

One potential idea for this that I think is inline with the design of other parts of the interface could be to have `s` do a regular split as it does today, and `S` could open up a menu with options like `p` for `split --parallel`.

## Describe alternatives you've considered
I could probably configure a custom command to do this using the `$checked_files` placeholder.

## Why current features fail to cover this request
Given that this isn't interacting with any other tools besides jujutsu, and it isn't a super niche workflow, I think it's reasonable to have this built-in to the existing split functionality rather than leaving it up to the user to create a custom command.

idursun:
Thanks! I will have a look.

PR #324, opened 2025-10-04
Add "SplitParallel" command

ianjsikes:
A potential implementation of #312 . Adds a new "SplitParallel" command that functions exactly the same as the existing "Split" command, except it passes the `--parallel` flag to `jj split` ([relevant jj docs](https://jj-vcs.github.io/jj/latest/cli-reference/#options_51)). I have it bound to `alt+s`, as `S` is already used by the squash mode.

I'm not sure if it would be a better approach to treat Split as its own mode. That would be more inline with some of the existing operations with multiple possible options, like Squash. I didn't attempt that in this PR because turning the existing single keypress Split into the entry point of a new mode may be a breaking change / may be disruptive of existing workflows. Feel free to close this PR if this solution isn't inline with the project's design.


internal/ui/operations/details/details.go
```diff
@@ -130,7 +130,23 @@ func (s *Operation) internalUpdate(msg tea.Msg) (*Operation, tea.Cmd) {
...
-					tea.Batch(s.context.RunInteractiveCommand(jj.Split(s.revision.GetChangeId(), selectedFiles), common.Refresh), common.Close),
+					tea.Batch(s.context.RunInteractiveCommand(jj.Split(s.revision.GetChangeId(), selectedFiles, false), common.Refresh), common.Close),
+					key.NewBinding(key.WithKeys("y"), key.WithHelp("y", "yes"))),
+				confirmation.WithOption("No",
+					confirmation.Close,
+					key.NewBinding(key.WithKeys("n", "esc"), key.WithHelp("n/esc", "no"))),
+			)
+			s.confirmation = model
+			return s, s.confirmation.Init()
+		case key.Matches(msg, s.keyMap.Details.SplitParallel):
```
idursun: Instead of duplicating this here, you can do something like:

```go
case key.Matches(msg, s.keyMap.Details.Split, s.keyMap.Details.SplitParallel):
isParallel := case key.Matches(msg, s.keyMap.Details.SplitParallel)
```
and pass `isParallel` down to `jj.Split`.

internal/ui/revisions/revisions.go
```diff
@@ -430,7 +430,10 @@ func (m *Model) internalUpdate(msg tea.Msg) (*Model, tea.Cmd) {
 				return m, m.op.Init()
 			case key.Matches(msg, m.keymap.Split):
 				currentRevision := m.SelectedRevision().GetChangeId()
-				return m, m.context.RunInteractiveCommand(jj.Split(currentRevision, []string{}), common.Refresh)
+				return m, m.context.RunInteractiveCommand(jj.Split(currentRevision, []string{}, false), common.Refresh)
+			case key.Matches(msg, m.keymap.SplitParallel):
```
idursun: Same approach in the previous comment can be done here too.

idursun (changes requested):
Thanks for working on this! There are some comments but overall looks good. Ready to go if you can address them.

All of this will be rewritten in 0.10.x series but I am happy to get this in and ship it in 0.9.4.

Issue #329, opened 2025-10-06
FR: add search for help menu page


baggiiiie:
## Is your feature request related to a problem? Please describe.
- For someone getting used to building muscle memory when using jjui, they always need to check help menu to see a keymap
- Instead of scanning through the help menu (frustrating!), it would be useful to search on the help menu to see their corresponding keymap. 
- e.g., the workflow would be: press `?` to open help menu, then hit `/` to search and filter jj commands

## Describe the solution you'd like
- a searching and/or filtering functionality like in `jjui` main page

idursun:
Hey thanks for the suggestion.

I agree. Help had only two columns at the beginning and grew into a giant with 3 columns now. I want to revisit the design of it but I am thinking of turning help into a command palette in the next version. I might salvage some of that work and port back to the current version.

baggiiiie:
help menu search is implemented, closing this issue!

PR #334, opened 2025-10-11
feat: add search functionality to help menu

baggiiiie:
## summary

**Add search/filter support for help page**

*see issue #329*

## feature

- added **search and filtering logic** to the help page  
  - if a `h.keyMap.*.Mode` matches the user’s search, the whole group of help menu entries under that mode is displayed  
  - if an individual entry matches, that entry and its associated `h.keyMap.*.Mode` are shown  
  - basically: you can now search by mode or by any specific entry keyword  

- implemented search bar input with `charmbracelet/bubbles/textinput`  

- introduced new struct types to organize help content  
  - **`helpItem`**: represents an individual help entry  
    - has `display` (what’s shown to user) and `searchTerm` (used internally for filtering)  
    - `searchTerm` combines keybind, name, and description so you can match across multiple fields  
  - **`itemGroup`**: groups related `helpItem`s under the same mode  
    - e.g. everything under *Rebase* mode goes into the `itemGroup` for rebase  
    - first entry of the group is treated as the header  

- extended the main tea `Model` with:
  - `defaultMenu` – the full help menu  
  - `filteredMenu` – current filtered state based on search  
  - `searchQuery` – user’s input string  

## fixes

- fixed wrong description for the `revert` key (was marked as “rebase”) — this was messing up search results. now correctly labeled as `revert`.  

## refactors

- moved help menu rendering logic into a separate file for better structure and readability  

## others / notes

- tried using `charmbracelet/bubbles/list` for the menu layout; didn’t work out since the menu uses 3 columns  
- current search logic is exact match only; **no fuzzy find yet**; plan to add fuzzy matching later  
- possible future UI polish: dimming + highlighting for search feedback  
  - unmatched `itemGroup`s could be dimmed  
  - matched ones stay normal color  
  - matched characters within items could be **highlighted**, similar to `fzf`  

## preview of help menu with search

image
image
image
image

### DIFF
--- a/internal/ui/helppage/help.go
+++ b/internal/ui/helppage/help.go
@@ -137,6 +137,7 @@ func filterList(column menuColumn, query string) menuColumn {
 		}
 
 		if len(matchedItems) > 1 {
+			matchedItems = append(matchedItems, helpItem{"", ""})
 			filtered = append(filtered, matchedItems)
 		}
 	}
--- a/internal/ui/helppage/help_item_groups.go
+++ b/internal/ui/helppage/help_item_groups.go
@@ -138,16 +138,19 @@ func (h *Model) buildMiddleGroups() menuColumn {
 			h.newBindingItem(h.keyMap.Details.Squash),
 			h.newBindingItem(h.keyMap.Details.Diff),
 			h.newBindingItem(h.keyMap.Details.RevisionsChangingFile),
+			helpItem{"", ""},
 		},
 		itemGroup{
 			h.newModeItem(&h.keyMap.Evolog.Mode, "Evolog"),
 			h.newBindingItem(h.keyMap.Evolog.Diff),
 			h.newBindingItem(h.keyMap.Evolog.Restore),
+			helpItem{"", ""},
 		},
 		itemGroup{
 			h.newModeItem(&h.keyMap.Squash.Mode, "Squash"),
 			h.newBindingItem(h.keyMap.Squash.KeepEmptied),
 			h.newBindingItem(h.keyMap.Squash.Interactive),
+			helpItem{"", ""},
 		},
 		itemGroup{
 			h.newModeItem(&h.keyMap.Revert.Mode, "Revert"),
@@ -161,6 +164,7 @@ func (h *Model) buildMiddleGroups() menuColumn {
 			h.newBindingItem(h.keyMap.Rebase.After),
 			h.newBindingItem(h.keyMap.Rebase.Onto),
 			h.newBindingItem(h.keyMap.Rebase.Insert),
+			helpItem{"", ""},
 		},
 		itemGroup{
 			h.newModeItem(&h.keyMap.Duplicate.Mode, "Duplicate"),
@@ -187,11 +191,13 @@ func (h *Model) buildRightGroups() menuColumn {
 			h.newBindingItem(h.keyMap.Preview.Expand),
 			h.newBindingItem(h.keyMap.Preview.Shrink),
 			h.newBindingItem(h.keyMap.Preview.ToggleBottom),
+			helpItem{"", ""},
 		},
 		itemGroup{
 			h.newModeItem(&h.keyMap.Git.Mode, "Git"),
 			h.newBindingItem(h.keyMap.Git.Push),
 			h.newBindingItem(h.keyMap.Git.Fetch),
+			helpItem{"", ""},
 		},
 		itemGroup{
 			h.newModeItem(&h.keyMap.Bookmark.Mode, "Bookmarks"),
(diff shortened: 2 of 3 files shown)


idursun:
Hey @baggiiiie thanks for working on this.

I have a couple of suggestions:

* Feel free to refactor the help page.
* I'd recommend it to be in search mode from the get go. User opens help and starts typing to filter it down, or presses `esc` to close it. I don't think there's need to press `/` to filter it down as the initial view is not scrollable anyway.
* I'd also like to see the mode names next to the bindings if I filter it down, because as you can see in the screenshot, I see a bunch of diff bindings but have no way to tell in which mode they can be used.
image


I'd probably try setting the entries to always belong to a mode/group by using something like this:

```golang
		h.modeEntry(h.keyMap.Rebase.Mode, "Rebase"),
			h.keyMap.Rebase.Revision,
			h.keyMap.Rebase.Source,
			h.keyMap.Rebase.Branch,
			h.keyMap.Rebase.Before,
			h.keyMap.Rebase.After,
			h.keyMap.Rebase.Onto,
			h.keyMap.Rebase.Insert,
		)
```
so that the each group/mode would filter down their own children. What do you think?

baggiiiie:
hey @idursun , thanks for the feedback, im working on a newer version of help menu with search, will push my changes soon for review!

baggiiiie:
hey @idursun i have pushed some changes and updated PR description. 
sorry for the big PR, i've documented as clearly and concisely as possible, together with git commit history. lemme know if there's anything else i could help to make this review easier!

internal/ui/helppage/help.go
```diff
@@ -52,162 +73,76 @@ func (h *Model) FullHelp() [][]key.Binding {
...
 	case tea.KeyMsg:
 		switch {
 		case key.Matches(msg, h.keyMap.Help), key.Matches(msg, h.keyMap.Cancel):
 			return h, common.Close
 		}
 	}
-	return h, nil
-}
 
-func (h *Model) printKeyBinding(k key.Binding) string {
```
baggiiiie, line 69: moved to `internal/ui/helppage/help_item_groups.go`

internal/ui/helppage/help.go
```diff
@@ -52,162 +73,76 @@ func (h *Model) FullHelp() [][]key.Binding {
...
-	keyAligned := fmt.Sprintf("%9s", key.Help().Key)
-	return lipgloss.JoinHorizontal(0, h.styles.shortcut.Render(keyAligned), h.styles.title.Render(name))
-}
+func (h *Model) filterMenu() {
+	query := strings.ToLower(strings.TrimSpace(h.searchQuery.Value()))
 
-func (h *Model) View() string {
-	var left []string
-	left = append(left,
-		h.printTitle("UI"),
```
baggiiiie, line 90: moved to `internal/ui/helppage/help_item_groups.go`

internal/ui/helppage/help.go
```diff
@@ -52,162 +73,76 @@ func (h *Model) FullHelp() [][]key.Binding {
...
-	)
-	return h.styles.border.Render(content)
+	h.filteredMenu = helpMenu{
+		leftList:   filterList(h.defaultMenu.leftList, query),
+		middleList: filterList(h.defaultMenu.middleList, query),
+		rightList:  filterList(h.defaultMenu.rightList, query),
+	}
 }
 
-func (h *Model) renderColumn(width int, height int, lines ...string) string {
```
baggiiiie, line 208: rendering logic moved to `internal/ui/helppage/help_menu_render.go`

internal/config/keys.go
```diff
@@ -49,7 +49,7 @@ func Convert(m KeyMappings[keys]) KeyMappings[key.Binding] {
 		ExecJJ:           key.NewBinding(key.WithKeys(m.ExecJJ...), key.WithHelp(JoinKeys(m.ExecJJ), "interactive jj")),
 		ExecShell:        key.NewBinding(key.WithKeys(m.ExecShell...), key.WithHelp(JoinKeys(m.ExecShell), "interactive shell command")),
 		Revert: revertModeKeys[key.Binding]{
-			Mode:   key.NewBinding(key.WithKeys(m.Revert.Mode...), key.WithHelp(JoinKeys(m.Revert.Mode), "rebase")),
+			Mode:   key.NewBinding(key.WithKeys(m.Revert.Mode...), key.WithHelp(JoinKeys(m.Revert.Mode), "revert")),
```
idursun, line 52: Thanks!

internal/ui/helppage/help.go
```diff
@@ -52,162 +73,76 @@ func (h *Model) FullHelp() [][]key.Binding {
 }
 
 func (h *Model) Init() tea.Cmd {
-	return nil
+	return textinput.Blink
 }
```
idursun, line 77: So, even though this makes sense, this is problematic because of how bubbletea simulates a visual cursor. 

This will send a `BlinkMsg` at every blink interval, which is fine when help window is showing. However, it will keep sending it even when the help page is closed which will make jjui rerender unnecessarily. This only happens if the model is using a pointer receiver as it is the case here.

There's no way you would know this. I have sunk too many hours debugging similar issues in the past eventually gave up and don't use Blink anymore.

idursun (approved):
Beautiful work! Thank you very much.

There's one comment about the use of the Blink command, which I will fix it myself after merging this.

internal/ui/helppage/help.go
baggiiiie: oh shoot, thanks for catching this!

Issue #337, opened 2025-10-15
FR: Abandon a Tree of Changes


orthros:
## Is your feature request related to a problem? Please describe.
My workflow will often leave large dangling "trees" of commits from the main branch.

For example

A -> B -> C -> D -> @
    -> E -> F -> G
            -> H -> I

And I will usually need to abandon Change E as well as all of E's descendants (F, G, H and I)

This can be done on the command line with `jj abandon E::` 

I would like similar functionality in jjui

## Describe the solution you'd like

Introduce a new command `prune` in `jjui` which takes a single revision and runs `jj abandon -r {revision}::`

This command will be quite similar to the current Abandon command.

Default keybinding for it will be `alt + a` to make it difficult to accidentally invoke.

The command will print a warning so that users are informed of what they are doing.

## Describe alternatives you've considered

The alternatives I've considered are implementing this myself in a bash script, or extending the current "abandon" functionality.

## Why current features fail to cover this request

Abandon does work with a selected range, but with complex trees and the requirement to select each Change, it can be labor intensive. It is simpler and nicer to simply select the "root" of the tree you wish to be abandoned and "prune" it from there

## Additional context

<None>

idursun:
Hey, thanks.

You can do this by defining a custom command like the following:

```toml
[custom_commands]
"prune" = { key = ["alt+a"], args = ["abandon", "-r", "$change_id::"] }
```

PR #338, opened 2025-10-15
feat(prune): adds new command `prune`

orthros:
Prune takes a selected change and abandons it as well as all of it's descendants.

Consider the following tree
```
@
|
A   I
|   |
B G H
| |/
D F
|/
E
|
J
```
Make change F selected and execute the `prune` command (by default the keybinding is `alt + a`). It will warn the user, asking for confirmation, and then run `jj abandon -r F::`. 

This will abandon changes F, G, H, and I.

See Issue #337

### DIFF
--- a/internal/config/default/config.toml
+++ b/internal/config/default/config.toml
@@ -14,6 +14,7 @@ limit = 0
   commit = ["c"]
   refresh = ["ctrl+r"]
   abandon = ["a"]
+  prune = ["alt+a"]
   diff = ["d"]
   quit = ["q"]
   help = ["?"]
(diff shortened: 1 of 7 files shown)


idursun:
Hey, thanks for working on this. 

I am not sure if you have seen my comment in the linked issue, but in short, you can already do this using custom commands.

Even if it was not possible with custom commands, I would still prefer existing `abandon` feature to be changed to support this rather than duplicating the operation. Secondly, I generally prefer the name of the action to match the command to run. So calling an action prune is going to confuse some of the users (including me) as they wouldn't know what to expect (and currently we don't have a mechanism to show the command to run).

PR #348, opened 2025-10-22
feat: auto-detect preview placement, on by default

lbonn:
It's usually beneficial to display preview at the bottom in a tall terminal.

With these changes, jjui will try to detect it automatically while still letting the user switch between horizontal and vertical splits.

tig behaves in a similar way.

### DIFF
--- a/internal/ui/ui.go
+++ b/internal/ui/ui.go
@@ -309,7 +309,7 @@ func (m Model) updateStatus() {
 
 func (m Model) UpdatePreviewPosition() {
 	if m.previewModel.AutoPosition() {
-		atBottom := m.Height >= m.Width / 2
+		atBottom := m.Height >= m.Width/2
 		m.previewModel.SetPosition(true, atBottom)
 	}
 }


internal/config/default/config.toml
```diff
@@ -132,7 +132,7 @@ limit = 0
   revision_command = ["show", "--color", "always", "-r", "$change_id"]
   oplog_command = ["op", "show", "$operation_id", "--color", "always"]
   file_command = ["diff", "--color", "always", "-r", "$change_id", "$file"]
-  show_at_bottom = false
+  position = "auto"
```
vic, line 135: How can people know possible valid values for position? Remember that most people don't tend to read documentation, perhaps adding a comment of possible values is enough ?

lbonn: Yes that's a good point. I think I'll try to follow an approach similar to SuggestMode, recently introduced

lbonn: @vic updated

Though log.Fatal does not show in stdout stderr by default. One has to define DEBUG=1 and look in debug.log https://github.com/lbonn/jjui/blob/6e95d19846d33eb0fbd07d9563a2ad02cd9f9136/cmd/jjui/main.go#L118

idursun (approved):
Thanks!

Issue #350, opened 2025-10-24
ace jump panic with multiple prefix matches


baggiiiie:
## Steps to Reproduce the Problem

1. `f` to activate ace jump
2. type a character that matches multiple prefixes
3. hit `enter` (jump to the first match as expected)
4. hit `f` again
5. panic, program crashes

## Expected Behavior

step 4: `f` activates ace jump

## Actual Behavior

step 4: `f` causes panic

see below screen recording:
[![asciicast](https://asciinema.org/a/PeLH7srhOTE4LOsK8HPqkrfSN.svg)](https://asciinema.org/a/PeLH7srhOTE4LOsK8HPqkrfSN)

## Specifications

- Platform: MacOS
- Version: 0.9.5

baggiiiie:
ah i think i found the issue, [here](https://github.com/idursun/jjui/blob/6e95d19846d33eb0fbd07d9563a2ad02cd9f9136/internal/ui/operations/ace_jump/operation.go#L118-L121) it sets aceJump to nil but didn't return a close message but a nil, the operation model didn't close properly

baggiiiie:
hi @idursun,

i just opened a PR [fix(ace_jump): close operation after applying jump](https://github.com/idursun/jjui/pull/351) to fix this issue, have a look when you can, thanks!

idursun:
Thanks, it is merged now.

PR #351, opened 2025-10-24
fix(ace_jump): close operation after applying jump

baggiiiie:
## summary 
**fix ace jump panic after model not being closed properly**.   

*bug details and reproducible steps see #350*

## changes
a pretty short and self explanatory PR

- return `common.Close` after `enter` is pressed so that operation model will be closed and reset upon next ace jump activation


vic:
:+1: lgtm!

idursun (approved):
Thanks!

Issue #295, opened 2025-09-02
Paths with spaces not escaped properly.


Jollywatt:
## Description

Paths with spaces do not get escaped properly when editing files in the file search view.

## Steps to Reproduce the Problem

1. Assume a repo with `file/Two Words.txt`
1. Press `ctrl+t` to open file search
1. Select a path with spaces and edit with `alt+e`

## Expected Behavior

The correct file is opened with an editor.

## Actual Behavior

The path is split at spaces and the editor opens a new file at the first it.

## Specifications

- Platform: macOS
- Version: v0.9.3

idursun:
Thanks for the report, I will have a look.

baggiiiie:
hey @Jollywatt 

im trying to reproduce, but im not sure if the expected behavior is to open the file in the editor.

when `alt+e` is used on a revision, it runs `jj edit -r $revision --ignore-immutable`. this itself won't open editor.

do you have any other configs?

idursun:
@baggiiiie It's this line that's problematic: https://github.com/idursun/jjui/blob/e5869d0cada3d4cc87ce2f9489b5d9c0ca5a9f96/internal/ui/fuzzy_files/fuzzy_files.go#L148

So, in the find files view (ctrl+t), you can press `alt+e` to edit the file, but if the file name contains spaces then the command ran is something like:

```
sh -c "nvim Two Words.txt"
```

the file name argument needs to be quoted for this to work.

baggiiiie:
@idursun ah okay got it, didn't know `alt+e` could be used on files during fzf. learning new stuff on jjui every day :)

idursun:
PR is merged. Thanks!

PR #354, opened 2025-10-25
fix(fuzzy_files): quote path for editor command

baggiiiie:
## changes
**quote file path to properly process path with space and other shell stuff**
*see #295* 

## implementation

- rather than quoting at `Line: config.GetDefaultEditor() + " " + path`, add quoting at `fuzzy_search.SelectedMatch()` to address all potential issues
https://github.com/idursun/jjui/blob/e5869d0cada3d4cc87ce2f9489b5d9c0ca5a9f96/internal/ui/fuzzy_files/fuzzy_files.go#L148

### DIFF
--- a/internal/ui/fuzzy_search/fuzzy_search.go
+++ b/internal/ui/fuzzy_search/fuzzy_search.go
@@ -1,6 +1,7 @@
 package fuzzy_search
 
 import (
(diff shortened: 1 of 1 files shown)


idursun (approved):
Thanks!

Issue #355, opened 2025-10-27
Failing to preview 'copied' files.


aids12321:
## Description

When `jj status` shows a file is copied (prefixed by `C`), jjui fails to show a preview of the file.

## Steps to Reproduce the Problem

1. Create an empty repo.
2. Commit one file with some text.
3. Copy the file, and make minor changes to both the new and original.

## Expected Behavior
When viewing the copied file preview in jjui, you should see the changes between the original file and it's current state.

## Actual Behavior
The preview window is empty.

## Specifications

- Platform: Arch Linux
- Version: jj: 0.34.0, jjui: 0.9.5

baggiiiie:
hi @aids12321 thanks for reporting this!

im looking into it, seems like `copied` status slipped out in the processing. 
https://github.com/idursun/jjui/blob/f7b614001e4388f066a2a3b7be10e5fbf8212fae/internal/ui/operations/details/details.go#L317-L324

sending a fix soon!

idursun:
Hey thanks for the report. My question is how one can get to "created" file state. Whenever I create a new file, it is shown as added. If you can help me to reproduce it on my end, then it is fairly easy fix.

idursun:
PR is merged!

PR #357, opened 2025-10-27
fix for preview copied status file and filename with {} 

baggiiiie:
## summary
**Fix preview for files with `C` copied status and detail view with filenames with `{}`**

*for `C` copied file preview, see #355*.

## cause
`C` copied status file preview:
- when the status for the file is `C`, jjui runs command similar to `jj diff --color always -r kq 'file:"{new_file.md => copied_file.md}"'`
- it doesn't parse the filename correctly as `C` status isn't registered
- which causes `jj` to fail with `Warning: No matching entries for paths: {new_file.md => copied_file.md}`

filename with `{}`:
- as filenames with status like `C`(copied) and `R`(renamed) contain `{}`, and [here](https://github.com/idursun/jjui/blob/f7b614001e4388f066a2a3b7be10e5fbf8212fae/internal/ui/operations/details/details.go#L329-L332) uses `strings.Contains` and `strings.Index` to match `{}`, it causes panic and index out of bound error when filename contains `{}`
- see reproduction of this bug: https://asciinema.org/a/4BiXzIguS86T9WLLLmliezYlw 

## fix
- added `C` status and relevant changes
- updated to use regex for filename matching
- added tests for filenames with `{}`
- added tests for directory level file operation (details see screenshots below):
```
R {src/new_file_3.md => new_file.md}  # moving out of directory src
R src/{new_file.py => renamed_py.py} # moving within directory src
R {src1/to_be_renamed.md => src2/renamed.md}  # moving from dir src1 to src2
```
- changed to `iota` enum for file `status`

## after fix
what `jj` shows:
image
what `jjui` shows:
image

### DIFF
--- a/internal/ui/operations/details/details_test.go
+++ b/internal/ui/operations/details/details_test.go
@@ -124,3 +124,50 @@ func TestModel_Update_HandlesMovedFiles(t *testing.T) {
 	tm.Quit()
 	tm.WaitFinished(t, teatest.WithFinalTimeout(3*time.Second))
 }
+
+func TestModel_Update_HandlesMovedFilesInDeepDirectories(t *testing.T) {
+	commandRunner := test.NewTestCommandRunner(t)
+	commandRunner.Expect(jj.Snapshot())
+	commandRunner.Expect(jj.Status(Revision)).SetOutput([]byte("false false false\nR {src/new_file_3.md => new_file.md}\nR src/{new_file.py => renamed_py.py}\nR {src1/to_be_renamed.md => src2/renamed.md}\n"))
+	commandRunner.Expect(jj.Restore(Revision, []string{"new_file.md", "src/renamed_py.py", "src2/renamed.md"}))
+	defer commandRunner.Verify()
+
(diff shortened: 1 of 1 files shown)


baggiiiie:
hey @idursun 

just saw your comments about how to recreate the bug in #355, this is what i have:

```bash
echo "copy1\ncopy2\ncopy3\n" > copy-source 

jj new 

# modify the source slightly and create a copy with similar content 
echo "copy1\ncopy2\ncopy3\nsource\n" > copy-source 
echo "copy1\ncopy2\ncopy3\ntarget\n" > copy-target

jj status
# show `C` prefix

jjui
# doesn't show `C`
```

idursun:
@baggiiiie FYI, you can add "fixes #issuenumber" in your PR description so that the issue is automatically closed when PR is merged.

baggiiiie:
> Very nice work 🫶

@idursun hey thank you for the encouraging words! love to contribute when i could! 


and thanks for letting me know about `fixes`, i wasn't aware of that at all
hilariously i've been using `see #issuenumber`, so close to use the magic keyword by accident lol

PR #365, opened 2025-11-12
Support interactive file-level restores

remo5000:
This helps with discarding chunks that shouldn't get included in the revision, e.g. if it's meant for debugging

https://github.com/user-attachments/assets/08f1ac0c-7514-4f0e-afc1-eabca0f3ad13


vic:
lgtm +1

baggiiiie (commented):
nice!

Issue #367, opened 2025-11-14
missing `--tracked` option for git fetch


baggiiiie:
currently in the `git` menu, there's `--tracked` option missing to pull only the tracked bookmarks from remotes

<img width="600" height="500" alt="Image" src="https://github.com/user-attachments/assets/536da82b-50a9-45fa-8f3c-5273fd240c0d" />

PR #368, opened 2025-11-14
ui/git: add --tracked to git fetch

baggiiiie:
## changes
- add git push option with `--tracked` flag

Fixes: #367


idursun (approved):
Thanks for the report and the fix!

PR #372, opened 2025-11-21
internal/parser: get revision ids with template prefixes

### LINKED ISSUE #228: Sometimes a bookmark name is detected as if it was commit_id
## Description

See the screenshot and comments on #226 

I noticed this for revisions that have a local bookmark. If a revision has a tracked remote bookmark everything works fine.

- create a revision with bookmarks
- try using ace jump
- sometimes the bookmark is highlighted as if being a commit_id

## Expected Behavior

Bookmark should not be part `row.Commit.CommitId`

## Actual Behavior

sometimes bookmark is CommitId

## Specifications

- Platform: Linux x86
- Version: #226 PR and most likely master [499fa83](https://github.com/idursun/jjui/commit/499fa83c51eb91752b3bd334d3229374e3e5424e)

### LINKED ISSUE #358: WIP: fix bookmarks being mistaken as CommitId
## issue to fix
- #228 

## root cause of bug
- current `CommitId` is identify with below logic
- which leads to any *hex like* string bookmarks like `ace` to be mistaken as `CommitId`
https://github.com/idursun/jjui/blob/360d61202da0d96669f2868e3754985c78b5dfa5/internal/parser/row_line.go#L31-L39
https://github.com/idursun/jjui/blob/360d61202da0d96669f2868e3754985c78b5dfa5/internal/parser/row.go#L199-L206

## What are you planning to implement?
- A different approach to get `CommitId` for `jj log` output

### approach 1 
- draft implementation see [here](https://github.com/idursun/jjui/commit/61dc99deef2ad84b0e74a2d3db3129eaf42fdfab#r169107360)
- Simplify logic to get CommitId
    - from my own experiments, `CommitId` is always the second last element in `GraphRowLine.Segments`
    - hence we could simply take the second last element of `GraphRowLine.Segments` as `CommitId`

- this is failing these two tests (see the tests removed in the [commit](https://github.com/idursun/jjui/commit/61dc99deef2ad84b0e74a2d3db3129eaf42fdfab#r169107360)):
    - Revision with extremely long bookmarks (according to the test, `CommitId` would be on a new line)
    - Revision is single line with description (unsure about this test)

- however, i wasn't able to recreate these two tests with `jjui`

### approach 2
- use templated `jj log` to get `CommitId`
    - with existing `jj` template, we could get `CommitId` and `ChangeId` **without** any complex string operation 
- below command works decently well to serve this purpose
```
jj log -T 'builtin_log_compact ++ change_id.shortest() ++ " " ++ commit_id.shortest()'
```
- which gives something like:
```
@  rwzrmrvk yc@yingchaoooo.com 13 minutes ago a3754fb0
│  (empty) (no description set)
│  r a    # this is the new line generated by the additional 
○  nlyyqkyr yc@yingchaoooo.com 13 minutes ago git_head() b1deea63
│  (empty) (no description set)
│  nl b
```

- note that: `│  r a  ` is the new line generated by the additional `change_id.shortest() ++ " " ++ commit_id.shortest()` template input
- we can easily get `ChangeId` and `CommitId` and not trim out the new line
- this approach a good strategy to get the IDs natively from `jj`, which is definitely very nice as it's error free!
- it takes a bigger refactor, but i think it's worth the work and will pay off long term if we have more functionalities depending on the IDs

## Breaking Changes
Would this implementation introduce any breaking changes? If so, describe them and any migration strategy.
- for now, only ace jump depends on `CommitId`

## Questions for Discussion
- approach 1 is a quick fix targeting the issue #228 
- approach 2 is a better way to address the root cause, use `jj`'s native response which is more consistent and error free than the original `isHexLike` string operation or getting second last element in *approach 1*

@idursun what do you think?
@vic would love to hear your thoughts too!

baggiiiie:
## summary

**use `jj log -T separate()` to reliably fetch revision ChangeID, CommitID, IsDivergent, replacing current "best guess" effort**

## changes

As proposed in #358, using `jj`'s native support (built-in `change_id/commit_id/divergent`) to get ChangeID and commitID would be more preferable than current "best guess" by checking log output sequence in order, which breaks when a bookmark is "HexLike" and "??".

The implementation of the proposal uses `jj log`'s `-T` template option, option for `-T` is `separate('',  $prefixes, $template)`, where `$prefixes` is `change_id.shortest(), commit_id.shortest(), divergent` and `$template` is the user-configured template, i.e., `config.Current.Revisions.Template`

This way, ChangeID, CommitID, and IsDivergent for a revision are retrieved *reliably*, and the prefixes are removed after prefix parsing before rendering to the ui.

Fixes #358
Fixes #228

## tests

- updated existing tests in `test/log_parser_test.go` to include `$change_id, $commit_id, $is_divergent` in test data

## others

- existing test `TestParser_Parse_NoCommitId` with `testdata/no-commit-id.log` has commit id, though the test name indicates otherwise, unsure what the purpose of this test is

### DIFF
--- a/internal/config/jj_config.go
+++ b/internal/config/jj_config.go
@@ -10,6 +10,9 @@ type JJConfig struct {
 	Revsets       struct {
 		Log string `toml:"log"`
 	} `toml:"revsets"`
+	Templates struct {
+		Log string `toml:"log"`
+	} `toml:"templates"`
 }
 
 func (c *JJConfig) GetApplicableColors() map[string]Color {
@@ -41,7 +44,6 @@ func parseConfig(configContent string) (*JJConfig, error) {
 	if err != nil {
 		return nil, err
 	}
-
 	return &config, nil
 }
 
--- a/internal/jj/commands.go
+++ b/internal/jj/commands.go
@@ -33,7 +33,7 @@ func ConfigListAll() CommandArgs {
 	return []string{"config", "list", "--color", "never", "--include-defaults", "--ignore-working-copy"}
 }
 
-func Log(revset string, limit int) CommandArgs {
+func Log(revset string, limit int, jjTemplate string) CommandArgs {
 	args := []string{"log", "--color", "always", "--quiet"}
 	if revset != "" {
 		args = append(args, "-r", revset)
@@ -41,12 +41,14 @@ func Log(revset string, limit int) CommandArgs {
 	if limit > 0 {
 		args = append(args, "--limit", strconv.Itoa(limit))
 	}
-	if config.Current.Revisions.Template != "" {
-		prefix := "change_id.shortest(), commit_id.shortest(), divergent"
-		// prefix := "format_short_id(change_id), format_short_id(commit_id), divergent"
-		template := fmt.Sprintf("separate('',  %s, %s)", prefix, config.Current.Revisions.Template)
-		args = append(args, "-T", template)
+	prefix := "change_id.shortest(), commit_id.shortest(), divergent"
+	template := config.Current.Revisions.Template
+	// If jjui's template is empty, fall back to jj's templates.log
+	if template == "" {
+		template = jjTemplate
 	}
+	template = fmt.Sprintf("separate('',  %s, %s)", prefix, template)
+	args = append(args, "-T", template)
 	return args
 }
 
@@ -331,7 +333,10 @@ func Duplicate(from SelectedRevisions, to string, target string) CommandArgs {
 }
 
 func Evolog(revision string) CommandArgs {
-	return []string{"evolog", "-r", revision, "--color", "always", "--quiet", "--ignore-working-copy"}
+	prefix := "commit.change_id().shortest(), commit.commit_id().shortest(), commit.divergent()"
+	template := "builtin_evolog_compact"
+	template = fmt.Sprintf("separate('',  %s, %s)", prefix, template)
+	return []string{"evolog", "-r", revision, "--color", "always", "--quiet", "--ignore-working-copy", "--template", template}
 }
 
 func Args(args ...string) CommandArgs {
--- a/internal/parser/log_parser.go
+++ b/internal/parser/log_parser.go
@@ -4,6 +4,8 @@ import (
 	"io"
 )
 
+// ParseRows reads all rows from the provided reader and returns them as a
+// slice. It is used for `jj log` and `jj evolog` commands
 func ParseRows(reader io.Reader) []Row {
 	var rows []Row
 	controlChan := make(chan ControlMsg)
--- a/internal/ui/graph/graph_streamer.go
+++ b/internal/ui/graph/graph_streamer.go
@@ -23,11 +23,13 @@ type GraphStreamer struct {
 	batchSize   int
 }
 
-func NewGraphStreamer(ctx appContext.CommandRunner, revset string) (*GraphStreamer, error) {
+// NewGraphStreamer runs `jj log` command with given revset and jjTemplate and
+// returns a GraphStreamer
+func NewGraphStreamer(ctx appContext.CommandRunner, revset string, jjTemplate string) (*GraphStreamer, error) {
 	streamerCtx, cancel := context.WithCancel(context.Background())
 	var commandError error
 
-	command, err := ctx.RunCommandStreaming(streamerCtx, jj.Log(revset, config.Current.Limit))
+	command, err := ctx.RunCommandStreaming(streamerCtx, jj.Log(revset, config.Current.Limit, jjTemplate))
 	if err != nil {
 		cancel()
 		return nil, err
@@ -78,6 +80,7 @@ func NewGraphStreamer(ctx appContext.CommandRunner, revset string) (*GraphStream
 		batchSize:   batchSize,
 	}, commandError
 }
+
 func (g *GraphStreamer) RequestMore() parser.RowBatch {
 	g.controlChan <- parser.RequestMore
 	return <-g.rowsChan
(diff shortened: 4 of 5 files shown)


internal/parser/streaming_log_parser.go
```diff
@@ -29,47 +29,40 @@ func ParseRowsStreaming(reader io.Reader, controlChannel <-chan ControlMsg, batc
...
 				if len(rows) > batchSize {
-					select {
-					case msg := <-controlChannel:
-						switch msg {
-						case Close:
-							return
-						case RequestMore:
-							rowsChan <- RowBatch{Rows: rows, HasMore: true}
-							rows = nil
-							break
```
baggiiiie, line 44: removed redundant break statement

test/log_parser_test.go
```diff
@@ -23,6 +23,8 @@ func TestParser_Parse_WorkingCopyCommit(t *testing.T) {
 	assert.True(t, row.Commit.IsWorkingCopy)
 }
 
+// TODO: what does no-commit-id mean?
+// `testdata/no-commit-id.log` has commit id
```
baggiiiie, line 27: unsure if this test serves its purpose, as it says `no commit id`, but commit id exists in `testdata/no-commit-id.log`

internal/parser/row.go
```diff
@@ -196,15 +189,6 @@ func isChangeIdLike(s string) bool {
...
 
-func isHexLike(s string) bool {
-	for _, r := range s {
-		if !unicode.Is(unicode.Hex_Digit, r) {
-			return false
-		}
-	}
-	return true
-}
-
```
baggiiiie, line 207: `isHexLike` becomes unused

vic:
haven't tested, but changes lgtm, this is awesome @baggiiiie :tada: :clap:

baggiiiie:
@vic thanks!
give it go too

idursun:
Hey @baggiiiie thanks for the work. I briefly skimmed the changes and pulled your branch and gave it a run and it looks like below and I don't think this is intentional
image

and looks like this with the short id format

image

baggiiiie:
@idursun oh wow i have been using this build for a few days and didn't bump into that issue at all
the expected output should look exactly the same as before this change

do you have a config for `config.Current.Revisions.Template`? and could you paste a image of the ui before this change for comparison?

idursun:
> do you have a config for config.Current.Revisions.Template?

No, that's for folks to override the default `jj` template. I don't have an override.

before: 
image

after:
image

baggiiiie:
@idursun hmmm i do not have that issue at all

hey @vic, would u be able to try out and see if you have any issue?

below screenshot and screen recording are captured with `'format_short_id(id)' = 'id.shortest()'` alias in `jj/config.toml`

- image   


<br> 

see the recording for a fresh clone from this branch & go build: 

[asciicast](https://asciinema.org/a/757350)

vic:
@baggiiiie, sure!

Looks similar here, I don't have `config.Current.Revisions.Template` set, but I do have [format_short_id alias](https://github.com/vic/vix/blob/den/modules/vic/jujutsu.nix#L69)

image

baggiiiie:
hey @vic thanks for getting back and providing your config!


unfortunately, i have no lucky reproducing the bug after hours, attempting with all following templates and any potentially suspicious ones in your config 
```
              default-command = [
                "log" 
                "--no-pager" 
                "--reversed" 
                "--stat" 
                "--template" 
                "builtin_log_compact_full_description"
                "--limit"
                "3"
              ];

            template-aliases = {
              "format_short_id(id)" = "id.shortest().upper()"; # default is shortest(12)
              "format_short_change_id(id)" = "format_short_id(id)";
              "format_short_signature(signature)" = "signature.email()";
              "format_timestamp(timestamp)" = "timestamp.ago()";
``` 



i pushed a new branch ([yc/parsing-debug](https://github.com/baggiiiie/jjui/commits/yc/parsing-debug/)) in my fork with relevant debug loggings and a fix for ace jump, it would be truly appreciated if you could run it with `DEBUG=1` and send over the `debug.log` 🙏🏼🙏🏼

vic:
@baggiiiie using `yc/parsing-debug` branch and `DEBUG=1` dumped this [debug.log](https://gist.github.com/vic/46c6a7dabce35d29c3f4b70ba5982aa9). Hope it is helpful.

idursun:
@baggiiiie here's my [debug.log](https://gist.github.com/idursun/e1c661606b59eb256be27268a75ee2ae)

baggiiiie:
@vic @idursun thank you guys for the help, much appreciated!! 🫶🏼🫶🏼

turned out it's my own config that's making things weird >: 

i have added a `Templates` field to [JJConfig](https://github.com/idursun/jjui/blob/1e2d9373e6cab15d39cf2a7c5d45644cd1be7d19/internal/config/jj_config.go#L7-L13), which is used to load `jj config list --include-defaults --ignore-working-copy`, and newly added `Templates` now saves `templates.log` and is used to run `Log` command

also, as ChangeID/CommitID are now fetched with `jj log` with `.shortest()`, they are always lowercase.
there might be a case where user uses uppercase for ui display (thanks @vic for your config!), and this breaks ace jump highlighting. this PR now has a fix for it.

all change details are in each commit's messages. thank yall again for helping out!

idursun:
Thanks @baggiiiie this is looking much better now. 

However, I noticed that `evolog` view is broken. It is using the the parser as revisions (the structure is pretty much the same) but I guess because `evolog` doesn't use a template now, the view is broken in the same way revisions was.

image

baggiiiie:
@idursun patched the same prefixes to evolog command: 
```
separate("",  commit.change_id().shortest(), commit.commit_id().shortest(), commit.divergent(), builtin_evolog_compact)
```

as jj now only has one default template for evolog, which is [builtin_evolog_compact](https://github.com/jj-vcs/jj/blob/c93682f218ea4e8a8a7aa4041618051f613a9d45/cli/src/config/templates.toml#L34), i didn't use any config to fetch `templates.evolog` from `jj config list`

PR #378, opened 2025-11-23
refactor: serialise command execution

### LINKED ISSUE #376: Lock command executor to prevent stale workspace situations
Currently, most of the commands executions are batched inside a `tea.Cmd` which potentially can run multiple commands around the same time, which can lead to stale workspaces if the commands snapshot. (e.g. auto refresh happening around the time user opens the details view). 

To prevent, and serialise the command execution, we need to add a sync mechanism to the `MainCommandRunner` implementation.

idursun:
This first attempt at serialising the command execution to prevent occasional stale workspace situations.

fixes #376

### DIFF
--- a/internal/ui/common/msgs.go
+++ b/internal/ui/common/msgs.go
@@ -27,8 +27,9 @@ type (
 	}
 	CommandRunningMsg   string
 	CommandCompletedMsg struct {
-		Output string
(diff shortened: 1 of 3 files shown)


idursun:
Seems to be working fine so I am merging this.

PR #384, opened 2025-11-25
refactor: introduce and implement common.Model

idursun:
We are almost always using pointer receiver models so we don't need to return the model as it never gets replaced. Additionally, bubbletea v2 is going to change the tea.Model and it is only expected to be implemented at the root level.

### DIFF
--- a/internal/ui/helppage/help_test.go
+++ b/internal/ui/helppage/help_test.go
@@ -7,6 +7,7 @@ import (
 
 	tea "github.com/charmbracelet/bubbletea"
 	"github.com/charmbracelet/lipgloss"
+	"github.com/stretchr/testify/assert"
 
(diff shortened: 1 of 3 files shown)


internal/ui/common/autocompletion/autocompletion.go
```diff
@@ -9,6 +9,8 @@ import (
 	"github.com/idursun/jjui/internal/ui/common"
 )
 
+var _ common.Model = (*AutoCompletionInput)(nil)
```
baggiiiie, line 12: is this forcing compiler to check `AutoCompletionInput` implements `common.Model`?

idursun: Yes, a dirty trick to move the failure to compile time rather than failing ay runtime or misbehaving.

PR #390, opened 2025-11-26
Preview panel shows whole commit diff instead of selected file's diff when terminal regains focus

abourget:
When the terminal loses focus and regains it, the preview panel refreshes but displays the entire commit diff instead of the diff for the previously selected file. The file selection in the details list is preserved correctly, but the preview content is not in sync.

### Root Cause:

Multiple components overwrite each other's selections during refresh operations. When `tea.FocusMsg` triggers `RefreshMsg{KeepSelections: true}`, both the revisions and details components process the refresh, but the revisions component calls `updateSelection()` which overwrites the file selection with a revision selection, causing the preview to use `RevisionCommand` instead of `FileCommand`.

### Solution:

1. Preview panel: Only refresh on `RefreshMsg` if we have a `SelectedFile` in context, preventing whole-diff display
2. Revisions component: Add `keepSelections` flag to skip `updateSelection()` calls when `KeepSelections: true`
3. Details component: Preserve cursor position (highlighted file) during refresh instead of defaulting to first file

This ensures that when the terminal regains focus, the file selection and preview content remain consistent, showing the correct file diff instead of the whole commit diff.

---

I'm relatively new to the codebase, let me know if this is the wrong way to approve the problem.

### DIFF
--- a/internal/ui/operations/details/details.go
+++ b/internal/ui/operations/details/details.go
@@ -66,26 +66,29 @@ func (s *Operation) Update(msg tea.Msg) tea.Cmd {
 		return s.load(s.revision.GetChangeId())
 	case updateCommitStatusMsg:
 		items := s.createListItems(msg.summary, msg.selectedFiles)
-		var selectionChangedCmd tea.Cmd
 		s.context.ClearCheckedItems(reflect.TypeFor[context.SelectedFile]())
-		if len(items) > 0 {
-			var first context.SelectedItem
-			for _, it := range items {
+
+		for _, it := range items {
+			if it.selected {
 				sel := context.SelectedFile{
 					ChangeId: s.revision.GetChangeId(),
 					CommitId: s.revision.CommitId,
 					File:     it.fileName,
 				}
-				if first == nil {
-					first = sel
-				}
-				if it.selected {
-					s.context.AddCheckedItem(sel)
-				}
+				s.context.AddCheckedItem(sel)
 			}
-			selectionChangedCmd = s.context.SetSelectedItem(first)
 		}
 		s.setItems(items)
+
+		// Set selection to current cursor position
+		var selectionChangedCmd tea.Cmd
+		if current := s.current(); current != nil {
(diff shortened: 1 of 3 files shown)


idursun:
Hey, thanks for the contribution! 

Let me have a look at what's really happening.

internal/ui/operations/details/details.go
```diff
@@ -56,6 +56,45 @@ func (s *Operation) Init() tea.Cmd {
 	return s.load(s.revision.GetChangeId())
 }
 
+func (s *Operation) preserveSelection(items []*item, currentHighlightedFile string) (context.SelectedItem, int) {
```
baggiiiie: im not sure if a whole function of `preserveSelection` is needed, since there's `context.AddCheckedItem` and `setItem`

baggiiiie:
@abourget i am hopelessly bad to express code with words, just [pushed a commit](https://github.com/baggiiiie/jjui/commit/f69a126db1b901497cf831d5edeb50df0fbf1918) to demo my ideas

it addresses the immediate issue (though im not sure if any edge cases are missing), feel free to have a look and play around and apply to your fix!

idursun:
I found out the underlying issue is, when refreshed, revisions send and updateSelection command which is causing the preview window to update its contents.

I have verified that details is actually keeping the cursor at where it should be. (same for evolog as it is also affected by this).

The fix should be that the revisions should run `updateSelection` if currently selected item is not a revisions. This should be enough to fix the details view problem but not the evolog problem (as it is working with RevisionItem's because evolog item is just the same change id with a different commit id). Having said that though, I see evolog fix can be done in a separate PR.

I agree with @baggiiiie points as the fix is a little bit more convoluted then it should be. I haven't done any work on this, so I don't know yet how feasible the approach I described above, but I feel that it should be straightforward for fixing the details view issue.

abourget:
ok, I've rebased and pushed a new proposition, taking your feedback into account. What do you think?

This one keeps the current line height too, so better than the previous too.

PR #391, opened 2025-11-26
feat: add basic mouse support

### LINKED ISSUE #110: FR: mouse wheel support / full mouse support
I'd love to be able to scroll the window with the mouse wheel, for the same reason as #109.

Additionally, but less importantly, I'm curious how hard it would be to be able to select commits and UI buttons with the mouse, and to allow a fully mouse-driven interface.

idursun:
This PR adds basic mouse support. Mouse support is only limited to the revisions view (excluding operations), preview and diff view.

This implementation is just to get the functionality in but actual mechanics subject to change later on. 

* Revisions view is scrollable and clickable. 
* Preview is draggable and scrollable.
* Diff view is scrollable

fixes #110 

https://github.com/user-attachments/assets/7cbad275-713e-4fa3-8d61-4e86c301ab0c

### DIFF
--- a/internal/ui/common/focusable.go
+++ b/internal/ui/common/focusable.go
@@ -23,6 +23,13 @@ type IMouseAware interface {
 	Scroll(delta int) tea.Cmd
 }
 
+type Draggable interface {
+	DragStart(x, y int) bool
+	DragMove(x, y int) tea.Cmd
+	DragEnd(x, y int) tea.Cmd
+	IsDragging() bool
+}
+
 type MouseAware struct {
 	dragging  bool
 	dragStart cellbuf.Position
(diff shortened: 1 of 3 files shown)


internal/ui/revisions/revisions.go
```diff
@@ -236,12 +298,24 @@ func (m *Model) Update(msg tea.Msg) tea.Cmd {
...
+		case tea.MouseActionPress:
+			if msg.Button == tea.MouseButtonLeft {
+				return m.ClickAt(msg.X, msg.Y)
+			} else if msg.Button == tea.MouseButtonWheelUp {
+				return m.Scroll(-3)
+			} else if msg.Button == tea.MouseButtonWheelDown {
+				return m.Scroll(3)
+			}
+			return nil
+		}
```
baggiiiie, line 316: since it's already in a switch case, we could keep case switching for consistency

```suggestion
                case tea.MouseActionPress:
                        switch msg.Button {
                        case tea.MouseButtonLeft:
                                return m.ClickAt(msg.X, msg.Y)
                        case tea.MouseButtonWheelUp:
                                return m.Scroll(-3)
                        case tea.MouseButtonWheelDown:
                                return m.Scroll(3)
                        }
                        return nil
                }

```

nickchomey:
REALLY nice work. 

One suggestion - change the mouse pointer when hovering the preview border, to indicate that it is draggable

PR #392, opened 2025-11-27
Coming back to previous state  when exec command failed

ArnaudBger:
This PR enhances the user experience when executing commands (jj or shell) by automatically restoring failed commands to the input field, allowing users to easily
edit and retry without having to retype the entire command.

### DIFF
--- a/internal/ui/status/status.go
+++ b/internal/ui/status/status.go
@@ -127,6 +127,20 @@ func (m *Model) Update(msg tea.Msg) tea.Cmd {
 		m.loadEditingSuggestions()
 		m.fuzzy, m.editStatus = fuzzy_files.NewModel(msg)
 		return tea.Batch(m.fuzzy.Init(), m.input.Focus())
+	case common.ExecProcessCompletedMsg:
+		if msg.Err != nil {
+			m.mode = "exec " + msg.Msg.Mode.Mode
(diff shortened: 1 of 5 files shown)


internal/ui/status/status.go
```diff
@@ -127,6 +127,20 @@ func (m *Model) Update(msg tea.Msg) tea.Cmd {
...
+		if msg.Err != nil {
+			m.mode = "exec " + msg.Msg.Mode.Mode
+			m.input.Prompt = msg.Msg.Mode.Prompt
+			m.loadEditingSuggestions()
+			m.fuzzy, m.editStatus = fuzzy_input.NewModel(&m.input, m.input.AvailableSuggestions())
+
+			// Avoid to change the current behavior when coming back from exec process
+			focusCmd := m.input.Focus()
+			keyMsg := tea.KeyMsg{Type: tea.KeyRunes, Runes: []rune(msg.Msg.Line)}
+			updateCmd := m.Update(keyMsg)
```
baggiiiie, line 140: maybe `m.input.SetValue` is more direct and clear than `m.Update` in this case, so no keyevent is needed

ArnaudBger: It was my original intention, but `m.input.SetValue` doesn't produce the exact same state. It results in unexpected behavior with the fuzzy_input model. Thanks for reviewing !

PR #395, opened 2025-11-28
list/renderer: fix viewport adjustment on PageDown

baggiiiie:
With the fix from #387, when PageDown moved the cursor below the viewport, the renderer broke early before rendering the focused item, so `selectedLineStart` and `selectedLineEnd` stayed -1, which leads to the viewport not being adjusted and rerendered.

Now the renderer continues rendering until it reaches the focused item, even if it's below the viewport. So that the viewport is adjusted to show the focused item and the screen scrolls to follow the cursor

broken:
[asciicast](https://asciinema.org/a/758652)

fix: 
[asciicast](https://asciinema.org/a/hW05tZYZm5jKPqoXYhxYg0zwD)

### DIFF
--- a/internal/ui/common/list/renderer.go
+++ b/internal/ui/common/list/renderer.go
@@ -109,7 +109,14 @@ func (r *ListRenderer) RenderWithOptions(opts RenderOptions) string {
 		if isFocused {
 			selectedLineEnd = r.totalLineCount()
 		}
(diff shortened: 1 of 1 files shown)


idursun (approved):
Thanks!

I think I am going to write a lot of tests to cover all these infractions

PR #399, opened 2025-11-29
revset: fix revset not using default when empty

baggiiiie:
When jjui is launched with a `--revset/-r` flag, the input argument should be come the default revset for the session, i.e., when user hits `L` and then `enter` with empty string, it should default to the argument defined by `-r` flag.

However, currently There's a mismatch between the display of `revset` and the actual jj log fetched: `revset` is user's session-wide default defined by `-r` flag, actual jj log is user's jj/jjui config.

This is due to `m.context.CurrentRevset` is updated to empty string, instead of default, when the text input is an empty string.

See 
image

[asciicast](https://asciinema.org/a/CJ0dsrGzlEWvv0ApI31GnHmAr)

### DIFF
--- a/internal/ui/revset/revset.go
+++ b/internal/ui/revset/revset.go
@@ -83,6 +83,10 @@ func New(context *appContext.MainContext) *Model {
 }
 
 func (m *Model) Init() tea.Cmd {
+	// Ensure CurrentRevset is initialized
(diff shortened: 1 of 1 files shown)


idursun (approved):
Thanks!

Issue #369, opened 2025-11-14
jjui duplicated description before `~`


baggiiiie:
## Steps to Reproduce the Problem

1. choose a revset that has `~` not at the bottom of screen
1. go to a revision before `~`
1. press `enter` to show description

## Expected Behavior
- description shows correctly

## Actual Behavior
- description somehow is duplicated
  
[![asciicast](https://asciinema.org/a/S0EuCys8xC1pt5pIuUZKGYGsf.svg)](https://asciinema.org/a/S0EuCys8xC1pt5pIuUZKGYGsf)

## Specifications

- Platform: MacOS
- Version: 0.9.5

idursun:
Thanks for the report! I’ll have a look.

PR #403, opened 2025-12-01
revision: fix double rendering of inline describe content

baggiiiie:
## [revision: fix double rendering of inline describe content](https://github.com/idursun/jjui/commit/060578f52a1e329c1484658a965b6131b6009cda)
Previously, inline describe content was rendered twice when the next
line in the revision list contains only a revision line (no separate
description lines below it)

Added a new tracking flag `revisionLineRendered` to properly sequence
the description overlay rendering, which tracks revision line rendering.

Following checks are made before rendering the description overlay:
- currently on a highlighted row AND
- revision line has already been rendered AND
- we need to render the description overlay AND
- we haven't already rendered it AND
- the current line is highlightable

Fixes https://github.com/idursun/jjui/issues/369

## [revision: break down item_renderer Render logic](https://github.com/idursun/jjui/commit/2f4620495b72ed6c1d1e161bd925109a93611cca)
revision: break down item_renderer Render logic
Refactored `Render` logic to smaller functions, separated into different
rendering logic, such as rendering for: 
- before revision line, 
- revision line,
- gutter
- etc

Also, removed unused function `aceJumpIndex`.

## others

the second commit is a bigger change with refactor
if the change list is too large, i could separate the two commits and submit individually

### DIFF
--- a/internal/ui/revisions/item_renderer.go
+++ b/internal/ui/revisions/item_renderer.go
@@ -31,8 +31,8 @@ type itemRenderer struct {
 
 func (ir itemRenderer) writeSection(w io.Writer, current parser.GraphGutter, extended parser.GraphGutter, highlight bool, section string, width int) {
 	isHighlighted := ir.isHighlighted
-	lines := strings.Split(section, "\n")
-	for _, sectionLine := range lines {
+	lines := strings.SplitSeq(section, "\n")
+	for sectionLine := range lines {
 		lw := strings.Builder{}
 		for _, segment := range current.Segments {
 			fmt.Fprint(&lw, segment.Style.Inherit(ir.textStyle).Render(segment.Text))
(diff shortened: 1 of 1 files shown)


idursun:
Hey, thanks! I'll have a look at this.

I couldn't reproduce it on my end but I assume you could and I am happy to see some tests. 🙌🏿

baggiiiie:
@idursun 

i wasn't too sure about this project's test philosophy since it's still being rapidly developed and refactored, but i've seen you are adding quite some tests lately, i'll help out and add tests when appropriate!

<br>

the bug happens when a **non-root** revision's parent in the graph is `elided` 

to reproduce:
- create a revision with multiple children (in below recording, the parent's change id is `u`)
- changes revset to `$ParentChangeID..` ( `u..` in this case)
- in each branch, parent node becomes `elided`
- press `enter` on any of the children revision of `u` 

[asciicast](https://asciinema.org/a/UeDh6Nt6cKNtCmBVHyQ37jydo)

idursun:
> i wasn't too sure about this project's test philosophy since it's still being rapidly developed and refactored, but i've seen you are adding quite some tests lately, i'll help out and add tests when appropriate!

The plan is to test as much as possible but writing tests was a lot of pain before. I find it easier to write tests now, therefore adding whenever I can.

PR #404, opened 2025-12-02
describe: catch cursor blinking to avoid unnecessary rendering

baggiiiie:
Currently, cursor.Blink msg is not being filtered out and it causes describe view to re-render unnecessarily and height is being calculated on every blinking. 

This change adds a filter to catch and ignore blinking messages.

## see below for debug message

https://github.com/user-attachments/assets/f5e47ecc-3c27-4690-b3ea-e726fef831c9

### DIFF
--- a/internal/ui/operations/describe/describe.go
+++ b/internal/ui/operations/describe/describe.go
@@ -1,6 +1,7 @@
 package describe
 
 import (
+	"github.com/charmbracelet/bubbles/cursor"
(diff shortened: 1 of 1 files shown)


baggiiiie:
> this will be a non issue when we start using bubbletea v2's native cursor support.

@idursun oh okay! 
i see some similar issues with other actions (like pressing arrow up/down at revision view), i'll go and see if it's something bubbletea v2 solves too

PR #405, opened 2025-12-03
operation: Add use destination message to squash operation

woutersmeenk:
Add the --use-destination-message option to the squash operation.

I added a extra toggle option to the squash operation:
image

Only after I had already implemented it did I read the 'discussion-first approach' for features. If needed I can open a issue first and change the implementation if needed, but since I already had the code I thought it was better to just open a PR.

### DIFF
--- a/internal/config/default/config.toml
+++ b/internal/config/default/config.toml
@@ -60,6 +60,7 @@ limit = 0
   [keys.squash]
     mode = ["S"]
     keep_emptied = ["e"]
+    use_destination_message = ["d"]
     interactive = ["i"]
   [keys.details]
     mode = ["l"]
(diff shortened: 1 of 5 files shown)


idursun:
Hey, thanks for the contribution.

I'd prefer an issue first but it is fine for this work as it is straight forward. However, if there was an issue, I'd recommend the following:

* If `--interactive` is set, then use `RunInteractiveCommand`
* else; if `--use-destination-message` then use`RunCommand` instead of the `RunInteractiveCommand` as no input from the user is required.
* If no arguments are set, then use `RunInteractiveCommand`

The idea is that we don't shell out if it is not needed.

woutersmeenk:
Thank you for you feedback. This makes sense indeed. I have implemented it.

idursun (approved):
Thanks!

PR #407, opened 2025-12-04
jj-update: fix commands to work with breaking changes in jj-0.36.0

baggiiiie:
in [Release v0.36.0 - jj-vcs/jj](https://github.com/jj-vcs/jj/releases/tag/v0.36.0), jj has breaking changes which breaks jjui. 

Below is the fixes for:
- deprecation of `--destination` flag: changed to `--onto`
- deprecation of `--edit` flag: changed to `--editor`
- deprecation of `--allow-new` flag for `git push`: removed all commands calling `git push` with `--allow-new`


internal/jj/bookmark_parser.go
```diff
@@ -24,10 +24,6 @@ type Bookmark struct {
 	CommitId  string
 }
 
-func (b Bookmark) IsPushable() bool {
-	return b.Local != nil && len(b.Remotes) == 0
-}
-
```
baggiiiie, line 30: `Bookmark.Pushable` now becomes used in the codebase 😔

idursun:
Hey, thanks!

If we merge this now, we will force everybody to be on 0.36, otherwise `jjui` won't work for them. Since 0.36 just came out, I would like to take advantage of the deprecation period for a little longer so that when this lands majority of the users would be on that version. I hope you don't mind this PR staying open for some time.

Additionally, we need to update the minimum jj version in the README.md to 0.36.

nickchomey:
I'm not sure anyone wins if this is merged or if you wait an arbitrary while.

Perhaps the fixes can be put behind some sort of flag/version check?

idursun:
> I'm not sure anyone wins if this is merged or if you wait an arbitrary while.

jj has been released just yesterday. I still want to keep my option to release another version of `jjui` that doesn't break for folks who are still on 0.35 or before. I sometimes update to the latest jj version a few weeks after its release.

> Perhaps the fixes can be put behind some sort of flag/version check?

Yes, but it's only worthwhile if implementing a version check and branching the code per jj version takes less than a couple of weeks.

baggiiiie:
> I hope you don't mind this PR staying open for some time.

@idursun hey it's fine, let's wait for a while 
also i have updated the readme to have minimum requirement of jj 0.36

> some sort of flag/version check?
handling backward compatibility is always a pain, since jj already has breaking changes, i think it's reasonable that jj's breaking changes propagate and become jjui's breaking changes

doprz:
> > I'm not sure anyone wins if this is merged or if you wait an arbitrary while.
> 
> jj has been released just yesterday. I still want to keep my option to release another version of `jjui` that doesn't break for folks who are still on 0.35 or before. I sometimes update to the latest jj version a few weeks after its release.
> 
> > Perhaps the fixes can be put behind some sort of flag/version check?
> 
> Yes, but it's only worthwhile if implementing a version check and branching the code per jj version takes less than a couple of weeks.

I'm using `jj v0.36.0` and although these commands are deprecated they still work albeit with a warning and what the correct syntax is moving forward. We could add a backwards compatibility flag or simply wait until jj makes a new release that removes these warnings/compatabilty layer in which then we would up jjui's minimum `jj` version to `v0.36.0+`.

nickchomey:
0.37.0 is now out. Looking through the changelog, not sure what affects jjui but there's a lot of welcomed changes.

https://github.com/jj-vcs/jj/releases/tag/v0.37.0

baggiiiie:
@nickchomey yeap didn't see anything that might affect jjui
i have upgraded to 0.37.0 and been using it for the whole day, didn't have any issue

nickchomey:
Are you running this PR branch then? Because not being able to use destination affects rebasing etc

baggiiiie:
> Because not being able to use destination affects rebasing etc

@nickchomey ah this should have been brought up in this PR, my bad!
since `--destination` is now `--onto` in jj 0.36.0+, i updated the keybinding for `-d` from `d` to `o`

@idursun please be aware of this behavior change, and advise if any changes needed

idursun (approved):
Thanks @baggiiiie I think this is good to go in.

PR #413, opened 2025-12-07
parser: stringify log/evolog prefixes to be color agnostic

### LINKED ISSUE #411: Corrupted change ID
## Description

The change ID detected by `jjui` is corrupted with part of the commit ID.

`~/.config/jj/config.toml`;

```toml
[colors]
change_id = "green"
commit_id = "green"
```

## Expected Behavior

`jjui` 0.9.6 and `jj log`:

```
@  unvwrskm (no email set) 2025-12-05 14:19:44 1d152460
│  (no description set)
```

## Actual Behavior

`jjui` 0.9.7;

```
@  unv1d15 (no email set) 2025-12-05 14:19:44 1d152460
│  (no description set)
```

## Specifications

- Platform: NixOS 25.11
- Version: 0.9.7

baggiiiie:
Previously, in PR #372, ChangeID/CommitID/IsDivergent is added to jj log/evolog commands to allow fetching these revision metadata natively with jj's template. 

However, the change didn't take into consideration of different color formatting of ChangeID/CommitID/IsDivergent and treated each of them as different `screen.Segment`s. 

When user configures them to be the same color,
ChangeID/CommitID/IsDivergent are actually returned as one chunk by jj and it becomes one `screen.Segment`. Hence the previous change breaks. 

This commit updates `ChangeID/CommitID/IsDivergent` prefixes with jj's `stringify` method, allowing them to be returned by jj consistently as one single chunk of text without color escape sequence. this way, `ChangeID/CommitID/IsDivergent` are always treated as one `screen.Segment` in jjui.

Also, `_PREFIX:` is added to the template for easier separation and parsing of prefixes.

Fixes #411


idursun (approved):
Thanks! 

I kicked the tires and seems to be working fine :shipit:

Issue #394, opened 2025-11-27
WIP: `/` search navigation and ace jump in rebase/squash/etc


baggiiiie:
## **What are you planning to implement?**

Here are two features that I have been using lately and enjoying a lot in my local jjui build:
- vim like search navigation (n/N) for `/`, implementation see [baggiiiie/jjui at yc/search-nav](https://github.com/baggiiiie/jjui/tree/yc/search-nav)
- ace jump for set_parents/duplicate/rebase/squash, implementation see [baggiiiie/jjui at yc/ace-jump](https://github.com/baggiiiie/jjui/tree/yc/ace-jump)

**NOTE**: 
- implementation are not finalized, just putting it out here to see if these are features people want
- im, at the same time, getting some mileage on them to see if there's any bugs

## **Proposed Approach**

Briefly describe your planned feature approach:
### search navigation

- search navigation (n/N), `n` for jumping forward and `N` backward
- highlighting on matched terms

### ace jump
User workflow:
1. Press 'r' to enter rebase mode (or 'S' for squash mode)
2. Press 'f' to activate ace jump while staying in rebase/squash mode
3. Type matching keys to jump to a revision
4. After jump completes, return to rebase/squash mode
5. Use existing rebase/squash keys to complete the operation

## **Testing Strategy**
How do you plan to test this implementation?
- Unit tests needed
- Manual testing approach

## **Breaking Changes**
Would this implementation introduce any breaking changes? If so, describe them and any migration strategy.
- No breaking changes

## **Questions for Discussion**

- any ideas/suggestions/feedback are welcome!

## **Additional Context**
Add any other context, mockups, or references that would help reviewers understand your proposal.

### search with highlighting and navigation

**any text on the screen is search-able!**

<img width="1006" height="365" alt="Image" src="https://github.com/user-attachments/assets/be367e3c-9d2f-42c9-88d6-9a91a6640498" />

### rebase/squash with ace jump
[![asciicast](https://asciinema.org/a/o1HZYIkxD74JRf6NkYmeuPTPH.svg)](https://asciinema.org/a/o1HZYIkxD74JRf6NkYmeuPTPH)

idursun:
> search navigation (n/N), n for jumping forward and N backward

There's already `'` for jumping to the next match, is your implementation going to different from what it does?

baggiiiie:
@idursun 
[...]

PR #414, opened 2025-12-07
revisions: add highlight to QuickSearch, make search case insensitive

baggiiiie:
## changes
revisions: add highlight to QuickSearch, make search case insensitive
- Added "quicksearch matched" style to dark/light themes 
- Added highlight to all case-insensitive matches of the search term
- Pressing enter/esc clears quicksearch and return to normal view

## others
This is a neat little feature that i have been using in my local build, as brought up in #394. 
did some polish on the code and opening this PR for visibility, would like to see if anyone else would enjoy this feature

## screenshot
image

### DIFF
--- a/internal/config/default/default_dark.toml
+++ b/internal/config/default/default_dark.toml
@@ -12,7 +12,7 @@ error = "red"
 "confirmation dimmed" = "white"
 "help title" = { fg = "green", bold = true }
 "revisions details selected" = { bg = "bright black" }
-"revisions matched" = { underline = false }
+"revisions matched" = { underline = false, reverse = true }
 "revset title" = "magenta"
 "revset text" = { fg = "green", bold = true }
 "revset completion text" = "white"
--- a/internal/ui/revisions/item_renderer.go
+++ b/internal/ui/revisions/item_renderer.go
@@ -20,6 +20,7 @@ type itemRenderer struct {
 	selectedStyle    lipgloss.Style
 	textStyle        lipgloss.Style
 	dimmedStyle      lipgloss.Style
+	matchedStyle     lipgloss.Style
 	isGutterInLane   func(lineIndex, segmentIndex int) bool
 	updateGutterText func(lineIndex, segmentIndex int, text string) string
 	inLane           bool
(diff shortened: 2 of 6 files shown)


internal/config/default/default_dark.toml
```diff
@@ -23,3 +23,4 @@ error = "red"
 "menu subtitle" = { fg = "230", bold = true }
 "menu matched" = { fg = "magenta", bold = true }
 "menu selected" = { fg = "cyan", bg = "default", bold = true, underline = false }
+"quicksearch matched" = { fg = "black", bg = "#FFB62E" }
```
idursun: can we use `revisions matched` here? Also, instead of defining an absolute background colour (which is very likely not translate well to different themes out there) we can just use `{ reverse = true }` which reversed fg and bg.

internal/ui/revisions/item_renderer.go
```diff
@@ -324,3 +331,30 @@ func (ir itemRenderer) Height() int {
 
 	return h
 }
+
+func (ir itemRenderer) renderQuickSearchHighlight(w io.Writer, textToRender string, style lipgloss.Style) {
+	searchTerm := ir.SearchText
+	lowercaseText := strings.ToLower(textToRender)
+
+	var lastIdx int
+	highlightStyle := common.DefaultPalette.Get("quicksearch matched")
```
idursun: We can pull this from the `itemRenderer`'s style values. You can define a `matchedStyle`.

internal/ui/revisions/revisions.go
```diff
@@ -282,8 +282,8 @@ func (m *Model) Update(msg tea.Msg) tea.Cmd {
 
 func (m *Model) internalUpdate(msg tea.Msg) tea.Cmd {
 	switch msg := msg.(type) {
-	//case rebase.updateHighlightedIdsMsg:
-	//	return m.op.Update(msg)
+	// case rebase.updateHighlightedIdsMsg:
+	// return m.op.Update(msg)
```
idursun: You can remove these lines. I think I accidentally pushed commented out code before.

baggiiiie:
@idursun nice thanks for the feedback!
i just pushed a commit with the changes

PR #415, opened 2025-12-08
feat: Lua scripting in custom commands

idursun:
> [!WARNING] 
> This is very much experimental.

This PR introduces the concept of `intent`s, which is my attempt to decouple capabilities from key bindings. This way input layer (i.e. keyboard, mouse, macros) translate user inputs to intents and only intents update the models.

I have implemented this only for the revisions, revset and flash models for now. I have consolidated all navigation operations into a single `Navigate` intent which can take various arguments for different kind of navigations/jumps. 

Additionally, I implemented a simple Lua script runner and a bridge for intents defined for the revisions, revset and flash models.

The following lua bridge defines the following functions:

  - revisions.current() → returns currently selected change ID (string) if a revision is selected.
  - revisions.checked() → returns a list of checked change IDs.
  - revisions.refresh({keep_selections?, selected_revision?}) → refreshes revisions 
  - revisions.navigate({by?, page?, target?, to?, fallback?, ensureView?, allowStream?}) → moves selection; supports delta navigation, page jumps, explicit targets (parent, child, working, etc.), and optional view/stream flags.
  - revisions.start_squash({files?}) → begins an inline squash workflow for the given files.
  - revisions.start_rebase({source?, target?}) → starts a rebase with parsed source/target strings (branch, descendants, after, before, etc.).
  - revisions.open_details() → opens the revision details view.
  - revisions.start_inline_describe() → opens inline describe input for the selected revision.
  - revset.set(value) → sets the current revset string.
  - revset.reset() → resets revset to default.
  - revset.current() → returns the active revset string.
  - revset.default() → returns the default revset string.
  - jj_async({...} | args...) → runs a JJ command asynchronously; yields until completion.
  - jj({...} | args...) → runs a JJ command immediately; returns (output, nil) on success or (nil, err) on failure.
  - flash(message) → enqueues a transient UI flash message.
  - copy_to_clipboard(text) → copies text to the system clipboard; returns (true, nil) on success or (nil, err) on failure.

All functions are available under the jjui table (e.g., jjui.revisions.refresh) and also exported at the global Lua scope for convenience (e.g., revisions.refresh).

I haven't fully tested anything but so far the following custom commands seem to work fine:

* Appends `| ancestors(<change id of the current revisions>, 2)` to the end of revset and bumps the number with each execution

```toml
[custom_commands.append_to_revset]
  key = ["+"]
  lua = '''
  local change_id = revisions.current()
  if not change_id then return end

  local current = revset.current()
  local bumped = false
  local updated = current:gsub("ancestors%(" .. change_id .. "%s*,%s*(%d+)%)", function(n)
    bumped = true
    return "ancestors(" .. change_id .. ", " .. (tonumber(n) + 1) .. ")"
  end, 1)

  if not bumped then
    updated = current .. "| ancestors(" .. change_id .. ", 2)"
  end

  revset.set(updated)
 '''
```

* Inserts a new commit after the selected one and then starts inline describe on the new revision.

```toml
[custom_commands.new_then_describe]
  key = ["N"]
  lua = '''
  jj("new", "-A", revisions.current())
  revisions.refresh()

  local new_change_id = jj("log", "-r", "@", "-T", "change_id.shortest()", "--no-graph")
  revisions.navigate{to=new_change_id}
  revisions.start_inline_describe()
  '''
```

* moves selection to `main`

```toml
[custom_commands.navigate_to_main]
  key = ["O"]
  lua = '''
  revisions.navigate{to=jj("log", "-r", "main", "-T", "change_id.shortest()", "--no-graph")}
  '''
```

* a copy to clipboard example

```toml
[custom_commands.copy_to_clipboard]
  key = ["X"]
  lua = '''
  local selections = revisions.checked()
  if #selections == 0 then
    flash("none selected")
  end
  local content = table.concat(selections, ",")
  copy_to_clipboard(content)
  '''
```

* a loop example

```toml
[custom_commands.loop]
  key = ["Y"]
  lua = '''
  local values = { "hello", "world" }
  for i = 1, #values do
    flash(values[i])
  end
  '''
```

### DIFF
--- a/internal/scripting/lua.go
+++ b/internal/scripting/lua.go
@@ -136,14 +136,14 @@ func (r *Runner) Done() bool {
 
 func registerAPI(L *lua.LState, runner *Runner) {
 	revisionsTable := L.NewTable()
-	revisionsTable.RawSetString("current_change_id", L.NewFunction(func(L *lua.LState) int {
+	revisionsTable.RawSetString("current", L.NewFunction(func(L *lua.LState) int {
 		if rev, ok := runner.ctx.SelectedItem.(uicontext.SelectedRevision); ok {
 			L.Push(lua.LString(rev.ChangeId))
 			return 1
 		}
 		return 0
 	}))
-	revisionsTable.RawSetString("selected_revisions", L.NewFunction(func(L *lua.LState) int {
+	revisionsTable.RawSetString("checked", L.NewFunction(func(L *lua.LState) int {
 		tbl := L.NewTable()
 		for _, item := range runner.ctx.CheckedItems {
 			if rev, ok := item.(uicontext.SelectedRevision); ok {
@@ -222,14 +222,11 @@ func registerAPI(L *lua.LState, runner *Runner) {
 		return 1
 	}))
 
-	root := L.NewTable()
-	root.RawSetString("revisions", revisionsTable)
-	root.RawSetString("revset", revsetTable)
-	root.RawSetString("run_async", L.NewFunction(func(L *lua.LState) int {
+	runAsyncFn := L.NewFunction(func(L *lua.LState) int {
 		args := argsFromLua(L)
 		return yieldStep(L, step{cmd: runner.ctx.RunCommand(args)})
-	}))
-	root.RawSetString("jj", L.NewFunction(func(L *lua.LState) int {
+	})
+	jjFn := L.NewFunction(func(L *lua.LState) int {
 		args := argsFromLua(L)
 		out, err := runner.ctx.RunCommandImmediate(args)
 		if err != nil {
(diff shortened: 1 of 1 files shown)


baggiiiie:
gosh this is looking great, amazing work!

idursun:
I am going to merge this now as this is mostly refactoring work and Lua script is more like an addition to the existing behaviour without any breaking changes.

abourget:
Do you plan to expose some elements of the `details` view? Like the selected files (absolute, relative), the position of the preview pane.. I'd like to have something that jumps into my editor at the exact same location as the preview pane :)

idursun:
@abourget That's the plan. I will add those gradually.

PR #420, opened 2025-12-11
feat: custom commands with sequence keys

idursun:
This PR adds `key_sequence` property to the custom commands so that they can be invoked by pressing multiple keys in sequence. It also adds a `desc` property which will come in handy later on.

```toml
[custom_commands.bookmark_list]
  key_sequence = ["w", "b", "l"]
  desc = "bookmarks list"
  lua = '''
  revset.set("bookmarks() | remote_bookmarks()")
  '''
```

Additionally, adds a sequence overlay which shows up when the user has pressed one of the first keys of a custom command that has `key_sequence` defined.

image

### DIFF
--- a/internal/ui/context/custom_command.go
+++ b/internal/ui/context/custom_command.go
@@ -11,14 +11,21 @@ import (
 
 type CustomCommand interface {
 	Binding() key.Binding
+	Sequence() []key.Binding
 	Description(ctx *MainContext) string
 	Prepare(ctx *MainContext) tea.Cmd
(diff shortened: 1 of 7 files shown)


vic:
neat! so we can drop leader keys now !?

EDIT: Please say YES. so we can have a unified custom commands feature.

idursun:
Hey, thanks! 

That's my plan :) I will brutally kill leader and custom commands in v0.10 and have only one way to customise `jjui`. 

Until then though, I am adding the building blocks to where they fit.

PR #421, opened 2025-12-12
main: pass through jj error

baggiiiie:
jjui doesn't pass through the stderr returned by jj, and errors are printed out as `Error: There is no jj repo in $DIR`
this change allows jj stderr to be caught and printed to user

## before
image

## after
image


baggiiiie:
> Thanks!

quest completed: fastest PR merge 😎

PR #422, opened 2025-12-12
feat(lua): add ability to await on operation results (cancelled/applied)

### LINKED ISSUE #310: FR: Inline commit
I like the `c` (commit) workflow, but I would prefer to write messages inline similar to the inline describe workflow.

Either:
- A new keyboard shortcut that starts the inline editor and when saved (using `alt+enter`) immediately starts a new commit afterwards.
- A new keyboard shortcut inside the existing inline editor, alongside the existing `alt+enter` shortcut. E.g. `alt+c` to save the inline description and then immediately start a new commit afterwards.

Of course I could just use inline describe and then press `n` immediately afterwards, but I like the idea of doing it all as one, so I don't forget and then end up accidentally adding more changes to the same commit.

N/A

## Additional context
N/A

idursun:
This PR updates the lua bridge to know if an operation has been applied or cancelled. This will allow #310 to be implemented like the following:

```toml
[custom_commands."inline commit"]
  key_sequence = ["w", "x"]
  lua = '''
  if revisions.start_inline_describe() then
    jj("new")
  end
  '''
```

A new revision is created only if the user pressed apply (i.e. `alt+enter`)

I’ll gradually apply the same approach to the remaining operations. 

fixes #310

### DIFF
--- a/internal/scripting/lua.go
+++ b/internal/scripting/lua.go
@@ -19,18 +19,19 @@ import (
 
 type step struct {
 	cmd     tea.Cmd
-	matcher func(tea.Msg) bool
+	matcher func(tea.Msg) (bool, []lua.LValue)
 }
 
 type Runner struct {
-	ctx     *uicontext.MainContext
-	main    *lua.LState
-	thread  *lua.LState
-	cancel  stdcontext.CancelFunc
-	fn      *lua.LFunction
-	started bool
-	await   func(tea.Msg) bool
-	done    bool
+	ctx        *uicontext.MainContext
+	main       *lua.LState
+	thread     *lua.LState
+	cancel     stdcontext.CancelFunc
+	fn         *lua.LFunction
+	started    bool
+	await      func(tea.Msg) (bool, []lua.LValue)
+	resumeArgs []lua.LValue
+	done       bool
 }
 
 func RunScript(ctx *uicontext.MainContext, src string) (*Runner, tea.Cmd, error) {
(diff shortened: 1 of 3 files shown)


OliverJAsh:
This is great @idursun 👏

I noticed your command doesn't move the cursor to the new commit. How would I go about doing that?

idursun:
You can move to any commit using:

```
revisions.navigate { to = <change id goes here> }
```

Having said that though, in my case, it is automatically moving to `@` already. Is it staying on the same revision and not moving to the new one in your case?

OliverJAsh:
I had to make a small change to make sure the new commit is inserted on top of the same commit that was just described (rather than the current working copy commit which might be a different commit):

```toml
[custom_commands."inline commit"]
  key_sequence = ["w", "x"]
  lua = '''
  if revisions.start_inline_describe() then
    jj("new", revisions.current())
  end
  '''
```

idursun:
Ah, I see. Then the following should work I think:

```toml
[custom_commands."inline commit"]
  key_sequence = ["w", "x"]
  lua = '''
  if revisions.start_inline_describe() then
    jj("new", revisions.current())
    revisions.navigate { to = "@" }
  end
  '''
  ```

**Comment by OliverJAsh** [comment]
When I run this on a selected commit that is not the working copy commit, the navigate call seems to interfere with the inline describe:

https://github.com/user-attachments/assets/918ab0b2-4789-4a03-9f9e-02aafa93d045

**Comment by idursun** [comment]
@OliverJAsh are you on the `main` branch or on 0.9.8? This only works on the main branch for now. I haven't released it yet.

I added refresh after `new` to correctly navigate to the new commit, seems to be working just fine on `main`.

```toml
[custom_commands."inline commit"]
  key_sequence = ["w", "x"]
  lua = '''
  if revisions.start_inline_describe() then
    jj("new", revisions.current())
    revisions.refresh()
    revisions.navigate { to = "@" }
  end
  '''
```

OliverJAsh:
I'm using 0.9.8, I will wait for the next release, thanks!

PR #423, opened 2025-12-13
Add option to hijack SSH Askpass to prompt for passphrase/pin

### LINKED ISSUE #100: Passphrase input for SSH remotes doesn't work?
When doing any operation on an SSH-based remote that would ask for a passphrase, `jjui` just shows the `  ⣾` animation in the bottom left and the command (eg. `jj git fetch --all-remotes`) and remains that way, but doesn't actually allow entering the passphrase. Pressing Enter a couple of times brings up the error:

```
 normal    ✗ jj git fetch --all-remotes
Error: Git process failed: External git program failed:
git@github.com: Permission denied (publickey).
fatal: Could not read from remote repository.

Please make sure you have the correct access rights
and the repository exists.
press  esc  to dismiss
```

which is the only way to exit from that state. Trying to enter the passphrase doesn't work, even Ctrl-C is ignored, pressing Enter and going to this failure mode seems the only way out. 

Is support for passphrase-based remote connections not there yet, or is this supposed to be working in some way?

oliverpool:
Fixes #100

As proposed in https://github.com/idursun/jjui/issues/100#issuecomment-3638654973, this adds a `[ssh] hijack_askpass` setting to prompt the user for the ssh passphrase/pin when needed.

The backend should works on linux and macos (platform support is limited by what https://github.com/tailscale/peercred supports) and is properly tested.

image

UI tasks:
- [x] display a the passphrase prompt and capture the user input
- [x] on `enter` return the captured password to the handler
- [x] hide the passphrase prompt when cancelled or done
- [x] override passphrase prompt on new request

### DIFF
--- a/internal/askpass/server.go
+++ b/internal/askpass/server.go
@@ -106,7 +106,8 @@ func (s *Server) smokeTestPeerCred(ln *net.UnixListener) error {
 		if err != nil {
 			return
 		}
-		conn.Close()
+		defer conn.Close()
+		_, _ = conn.Read(make([]byte, 1)) // darwin needs an active connection for peercred
 	}()
 	conn, err := ln.AcceptUnix()
 	if err != nil {
@@ -261,7 +262,7 @@ func ensureConnFromDescendant(conn *net.UnixConn, parentPID int) error {
 	}
 
 	for pid != parentPID {
-		pid, err = readPPid(pid)
+		pid, err = getPPid(pid)
 		if err != nil {
 			return err
 		}
@@ -271,22 +272,4 @@ func ensureConnFromDescendant(conn *net.UnixConn, parentPID int) error {
 		}
 	}
 	return nil
-}
-
-func readPPid(pid int) (int, error) {
-	f, err := os.Open(fmt.Sprintf("/proc/%d/status", pid))
-	if err != nil {
-		return 0, err
-	}
-	defer f.Close()
-	scan := bufio.NewScanner(f)
-	for scan.Scan() {
-		prefix, suffix, ok := strings.Cut(scan.Text(), "\t")
-		if !ok || prefix != "PPid:" {
-			continue
-		}
-		return strconv.Atoi(suffix)
-	}
-
-	return 0, scan.Err()
 }
--- a/go.mod
+++ b/go.mod
@@ -39,7 +39,7 @@ require (
 	github.com/rivo/uniseg v0.4.7
 	github.com/tailscale/peercred v0.0.0-20250107143737-35a0c7bd7edc
 	golang.org/x/crypto v0.46.0
-	golang.org/x/sys v0.39.0 // indirect
+	golang.org/x/sys v0.39.0
 	golang.org/x/text v0.32.0 // indirect
 	gopkg.in/yaml.v3 v3.0.1 // indirect
 )
--- a/internal/askpass/ppid_darwin.go
+++ b/internal/askpass/ppid_darwin.go
@@ -1,0 +1,13 @@
+package askpass
+
+import (
+	"golang.org/x/sys/unix"
+)
+
+func getPPid(pid int) (int, error) {
+	kproc, err := unix.SysctlKinfoProc("kern.proc.pid", pid)
+	if err != nil {
+		return 0, err
+	}
+	return int(kproc.Eproc.Ppid), nil
+}
--- a/internal/askpass/ppid_linux.go
+++ b/internal/askpass/ppid_linux.go
@@ -1,0 +1,27 @@
+package askpass
+
+import (
+	"bufio"
+	"fmt"
+	"os"
+	"strconv"
+	"strings"
+)
+
+func getPPid(pid int) (int, error) {
+	f, err := os.Open(fmt.Sprintf("/proc/%d/status", pid))
+	if err != nil {
+		return 0, err
+	}
+	defer f.Close()
+	scan := bufio.NewScanner(f)
+	for scan.Scan() {
+		prefix, suffix, ok := strings.Cut(scan.Text(), "\t")
+		if !ok || prefix != "PPid:" {
+			continue
+		}
+		return strconv.Atoi(suffix)
+	}
+
+	return 0, scan.Err()
+}
--- a/internal/askpass/ppid_other.go
+++ b/internal/askpass/ppid_other.go
@@ -1,0 +1,11 @@
+//go:build !darwin && !linux
+
+package askpass
+
+import (
+	"github.com/tailscale/peercred"
+)
+
+func getPPid(pid int) (int, error) {
+	return 0, peercred.ErrNotImplemented
+}


cmd/jjui/main.go
```diff
@@ -197,6 +204,43 @@ func main() {
...
+				})
+				return ""
+			case pw := <-password:
+				// TODO: send a message to discard the prompt
+				p.Send(common.CommandCompletedMsg{
+					Output: "askpass reply",
+				})
+				return pw
+			}
+		})
```
oliverpool: The remaining todos.

internal/askpass/server.go
```diff
@@ -0,0 +1,286 @@
...
+}
+
+type subprocess struct {
+	name    string
+	started chan struct{}
+	pid     *int
+	done    <-chan struct{}
+}
+
+func ensureConnFromDescendant(conn net.Conn, parentPID int) error {
```
oliverpool: As a security measure, only descendants from already known PIDs can prompt the user for a password (since any other process could connect to the socket and send a password-retrieving message).

cmd/jjui/main.go
```diff
@@ -197,6 +204,43 @@ func main() {
 	appContext.CurrentRevset = appContext.DefaultRevset
 
 	p := tea.NewProgram(ui.New(appContext), tea.WithAltScreen(), tea.WithReportFocus(), tea.WithMouseCellMotion())
+	if config.Current.Ssh.HijackAskpass {
+		if err := askpassServer.StartListening(); err != nil {
+			fmt.Fprintf(os.Stderr, "Error: ssh.hijack_askpass: %v\n", err)
+			os.Exit(1)
+		}
+		log.Println("askpass started")
+		defer askpassServer.Close()
```
oliverpool: btw the `os.Exit(1)` of the `main` don't play well with `defer`.

I would recommend wrapping `main()` inside a `run() error (or int)` to circumvent this footgun.

internal/ui/context/command_runner.go
```diff
@@ -66,14 +69,24 @@ func (a *MainCommandRunner) RunCommand(args []string, continuations ...tea.Cmd)
 	commands := make([]tea.Cmd, 0)
 	commands = append(commands,
 		func() tea.Msg {
+			started, cancel, env := a.Askpass.NewSubprocess(strings.Join(args, " "))
+			defer cancel()
```
idursun, line 74: As far as I can see, this will create a subprocess for every command we run, and depending on the feature we might run a lot of commands in parallel (e.g. bookmarks run two commands that loads up the moveable bookmarks and all bookmarks around the same time).

So I am not sure about the implications of this but doesn't seem ideal to me.

internal/askpass/server_test.go
```diff
@@ -0,0 +1,191 @@
...
+	started, cancel, env := s.NewSubprocess("ssh")
+	defer cancel()
+
+	cmd := exec.Command("ssh", "git@"+host, "-p", port,
+		"-o", "StrictHostKeyChecking=yes",
+		"-o", "UserKnownHostsFile="+knownHost,
+		"-o", "PubkeyAuthentication=no",
+		"-o", "PreferredAuthentications=password",
+		"-o", "PKCS11Provider=none",
+	)
```
idursun, line 129: I tried running the tests through `nix` and this failed. This will fail on every system that doesn't have `ssh` available, including package builds, possible windows builds as well.

internal/askpass/server.go
```diff
@@ -0,0 +1,286 @@
...
+		if pid == 0 {
+			pid, _ = cred.PID()
+			return fmt.Errorf("PID %d is not a descendant of PID %d", pid, parentPID)
+		}
+	}
+	return nil
+}
+
+func readPPid(pid int) (int, error) {
+	f, err := os.Open(fmt.Sprintf("/proc/%d/status", pid))
```
idursun: I think this is `unix` specific, will probably fail on other systems.

internal/ui/context/command_runner.go
oliverpool: `Askpass.NewSubprocess` is only adding a new entry in a map (so that if the socket receives a new connection, it can know from which command this came). It does not start a subprocess (only declares that a subprocess is about to be started).

So it shouldn't cause much burden (if the hijack is disabled, this is a no-op).

internal/askpass/server_test.go
oliverpool: Good point. I added an error check: if `ssh` client is not available, this test is skipped.

internal/askpass/server.go
oliverpool: That's why askpass performs a smoke test in `StartListening`: `smokeTestPeerCred`. It checks that `peercred` and `readPPid` work for a simple case.

oliverpool:
Thanks for the quick review!

I managed to hack a password prompt UI within the codebase. This can be tested by uncommenting a line in `main.go` (the result is discarded: it only allows to test the UI).

---

In my case I only need the password prompt when performing remote operations.

I think that for commit signing, you only need the private key (i.e. the passphrase), when writing (checking the signature only needs access to the public key, which should not require the passphrase).

_my last push (with ssh pin prompt) happened using `jjui` :tada:_

vic:
> _my last pushed happened using `jjui` 🎉_

we should all sign our commits with `-- pushed from my jjui` haha, like the old marketing strategy "sent from my iphone" to make more people look for jjui :D

oliverpool:
> Is there a path where we can check if a passphrase is needed at the start and don't care about it for the rest of the app?

This wouldn't address my usecase of a tpm-based ssh key. Each usage prompts for the PIN.

cmd/jjui/main.go
idursun: Yes, this makes sense. Can you move it to the a wrapper as a part of this PR? Basically does `os.Exit(run())`.

Then, `defer askpassServer.Close()` can be moved right after it is created.

idursun:
Hey, thanks! I am glad you could implement the UI part as well. Very much how I would implement it myself.

Please go ahead with this approach, as long as everything is behind a feature flag and no-op when the flag is off then I am happy to review it one more time once it is ready for review and merge.

Tests will fail because nix vendor hash won't match, I can fix it for you.

baggiiiie:
> we should all sign our commits with -- pushed from my jjui haha, like the old marketing strategy "sent from my iphone" to make more people look for jjui :D

i second this 😆 🫡

cmd/jjui/main.go
```diff
@@ -197,8 +207,48 @@ func main() {
...
+		// ensure that the prompt is not only made of spaces
+		for _, r := range s {
+			if !unicode.IsSpace(r) {
+				return s
+			}
+		}
+		return "ssh-askpass: "
+	}
+	return func(name, prompt string, done <-chan struct{}) (string, bool) {
+		password := make(chan string, 1)
```
baggiiiie: i think using `string` might not be a good practice for password, since it's immutable.

maybe use `[]byte` and explicitly cleared the byte slice after use

internal/ui/ui.go
```diff
@@ -364,6 +370,15 @@ func (m *Model) Update(msg tea.Msg) tea.Cmd {
 		m.previewModel.SetVisible(bool(msg))
 		cmds = append(cmds, common.SelectionChanged)
 		return tea.Batch(cmds...)
+	case common.TogglePasswordMsg:
+		if msg.Password == nil {
+			m.password = nil
+		} else if m.password != nil {
+			// todo: check conflict
```
baggiiiie: maybe just reject new password when `m.password` is not nil ?

oliverpool: This is now properly handled in https://github.com/idursun/jjui/pull/423/commits/a034b8e9ef7479f7dfb724621ed064c44b64fb2c
I tested manually with an `...-sk` key, which prompts the user to confirm on the device. Rejecting there triggers a new prompt which must override the current prompt.

cmd/jjui/main.go
oliverpool: Done in https://github.com/idursun/jjui/pull/423/commits/d6f29664c8dcf34659f157b9c38e34a3ba7b7be7

cmd/jjui/main.go
oliverpool: done in https://github.com/idursun/jjui/pull/423/commits/c2cfeb450c524e257ef47a817ddda81c72f699cd

internal/ui/context/command_runner.go
```diff
@@ -66,18 +70,32 @@ func (a *MainCommandRunner) RunCommand(args []string, continuations ...tea.Cmd)
...
+			started(c.Process.Pid)
+
+			err := c.Wait()
 			if err != nil {
 				var exitError *exec.ExitError
 				if errors.As(err, &exitError) {
-					err = errors.New(output.String())
+					msg := output.String()
+					if len(env) == 0 && slices.Contains([]string{"linux", "darwin"}, runtime.GOOS) {
+						msg += "\nHint: enable ssh.hijack_askpass if you expected a password prompt (e.g. ssh passphrase)"
```
oliverpool, line 96: I added a "just-in-time" hint: if a command fails and `ssh.hijack_askpass` is not enabled, inform the user about the existence of this setting.

The goal is to not clutter the interface for normal use (having an error here is unusual), while giving the user a chance to fix its setting quickly (without having to dig too deep in the docs).

idursun:
Hey thanks for the updates.

Test suite is failing on my mac. I don't know why though. I have `ssh` but that's about it.

In any case, can we scope these tests to `linux` only?

```
?       github.com/idursun/jjui/cmd/jjui        [no test files]
--- FAIL: TestServerListening (0.00s)
    server_test.go:56: unix.GetsockoptInt: socket is not connected
--- FAIL: TestServerAskpass (0.02s)
    server_test.go:103: unix.GetsockoptInt: socket is not connected
FAIL
FAIL    github.com/idursun/jjui/internal/askpass        1.320s
```

internal/ui/password/password.go
```diff
@@ -0,0 +1,81 @@
+package password
```
idursun, line 1: Can we rename this package to match the feature `askpass`. Folks might think this is a general purpose password.

oliverpool: The askpass package has only non-ui related stuff. 

Using the same name for this package might be confusing. What about `passwordpopup`? (This package is not specific to askpass)

idursun: Oh in that case `password` is fine. I thought for a second that it had dependency on the askpas package.

oliverpool:
>the tests are failing on my machine (Mac)

Indeed, macos needed some minor adjustments. Since I don't have a mac, I enabled the CI tests for macos and iterated until they passed: https://github.com/oliverpool/jjui/pull/1

They should now succeed on your computer.

internal/askpass/server.go
oliverpool: partially addressed in 96948b04478170bea07716c2387456244e3dea8c

PR #426, opened 2025-12-14
feat(nix): add comprehensive nix flake

doprz:
Features:

- [x] Nix Flake Configuration: Complete flake setup supporting multiple architectures
  - [x] flake-parts
  - [x] flake-parts [partitions](https://flake.parts/options/flake-parts-partitions.html)
  - [x] official [flake-compat](https://github.com/NixOS/flake-compat) (rev/hash pinned)
  - [x] `lib.fileset.unions` for `pkgs.buildGoModule`
  - [x] `pkgs.writeShellApplication` for vendorHash as an app
  - [x] Split `flake.nix` into modules
  - [x] Create dedicated `jjui` package with override support
  - [x] `nixpkgs.lib.systems.flakeExposed`
  - [x] Add `JJUI_CONF_DIR` env var notice/warning
  - [x] treefmt-nix
    - [x] nixfmt-rfc-style
    - [x] gofmt
- [x] Development Shell: Pre-configured dev environment
- [x] Multiple Entry Points: Support for `nix develop`, `nix run`, and `nix build`
- [x] Updated `.gitignore`
- [x] Added `.envrc` for nix flake users
- [x] Update `README.md`
- [x] Add `CODEOWNERS` file

TODO:
- [x] flake-compat
> I'm +1 on having support for stable-nix default.nix and shell.nix, however I don't like the fact that the PR adds flake-compat dependency for all flake users. flake-compat is NOT needed when using unstable-nix flakes. It is an anti-pattern of usage, because you will be introducing and forcing all fakes depending on this one to fetch flake-compat which is absolutely bloat for most.

> You can still use flake-compat in default.nix and shell.nix like this: https://github.com/vic/den/blob/d114d8b8d0f4c809a7057a7cef35e6d8648dcdaf/nix/template-packages.nix#L5 Just use a fixed rev and sha.

### DIFF
--- a/nix/package.nix
+++ b/nix/package.nix
@@ -19,7 +19,7 @@ buildGoModule {
     ];
   };
   vendorHash = lib.strings.trim (builtins.readFile ./vendor-hash);
-  doCheck = false;
+  doCheck = true;
 
   ldflags = [
     "-s"


idursun:
Hey, thanks! 

The only thing I can review in this PR is the .gitignore file and it looks good. Everything else is too much nix for me. 

Maybe @Adda0 can have a look as he is maintaining the nix builds.

Adda0:
Heya. Sure. While I am by no means an expert on the Nix ecosystem, I will have a look. Looks reasonable on a first glance.

doprz:
Thank you for the reviews and for the development of `jjui`, it's a great TUI that I've been daily driving!

idursun:
I am glad this is getting rid of the `vendor-hash`, which was a minor pain for me after updating dependencies.

flake.nix
```diff
@@ -1,9 +1,95 @@
 {
+  description = "A Text User Interface (TUI) designed for interacting with the Jujutsu version control system";
+
   inputs = {
+    nixpkgs.url = "github:NixOS/nixpkgs/nixos-unstable";
     systems.url = "github:nix-systems/default";
-    nixpkgs.url = "github:nixos/nixpkgs/nixpkgs-unstable";
     flake-parts.url = "github:hercules-ci/flake-parts";
+    flake-compat.url = "https://flakehub.com/f/edolstra/flake-compat/1.tar.gz";
```
vic: Please change this to be from NixOS/flake-compat, since it is the official flake-compat now.

Also, why this particular input uses flakehub.com ? change it to github like all others.

flake.nix
```diff
@@ -1,9 +1,95 @@
...
+    treefmt-nix.url = "github:numtide/treefmt-nix";
   };
 
-  outputs = inputs: inputs.flake-parts.lib.mkFlake { inherit inputs; } ./nix;
```
vic, line 8: I'd pretty much prefer not to have a monolithic flake.nix (see [reasons](https://x.com/oeiuwq/status/1999991219858854166?s=20)) if we are moving to a monolithic flake.nix then remove flake-parts.

Using flake-parts you could have `./nix/treefmt.nix` that is focused on configuring how to format the project, `./nix/shells.nix` for dev-shells, etc.

vic:
I'm +1 on having support for stable-nix `default.nix` and `shell.nix`, however I don't like the fact that the PR adds flake-compat dependency for all flake users. flake-compat is NOT needed when using unstable-nix flakes. It is an anti-pattern of usage, because you will be introducing and forcing all fakes depending on this one to fetch flake-compat which is absolutely bloat for most. 

You can still use flake-compat in default.nix and shell.nix like this: https://github.com/vic/den/blob/d114d8b8d0f4c809a7057a7cef35e6d8648dcdaf/nix/template-packages.nix#L5  Just use a fixed rev and sha.

vic:
> I am glad this is getting rid of the `vendor-hash`, which was a minor pain for me after updating dependencies.

It is still [there](https://github.com/idursun/jjui/pull/426/changes#diff-206b9ce276ab5971a2489d75eb1b12999d4bf3843b7988cbe8d687cfde61dea0R41), not just in a file which was used to keep track of version changes on a single file. you'd still have to make vendorHash value empty and build to get the new vendored hash when adding dependencies.

flake.nix
vic: If we are keeping flake-parts, it would be much better if your PR changes to use [partitions](https://flake.parts/options/flake-parts-partitions.html) so dependencies you've added like treefmt-nix is also not propagated to other flakes depending on jjui. The reason is treefmt is used for development of this repo only, and it should not be propagated.

vic:
Now that we are talking about dependencies, you could also remove `systems` input, and use `nixpkgs.lib.systems.flakeExposed` so jjui works on more systems and not only at [these](https://github.com/nix-systems/default/blob/main/default.nix)

.envrc
```diff
@@ -0,0 +1 @@
+use flake
```
vic, line 1: In my development shell I also export JJUI_CONF_DIR to something gitignored inside my development jjui repo, because that allows you to test jjui configurations in development working-copy. that way you don't have to change your system config to test stuff. 

Maybe we could have a comment about setting JJUI_CONF_DIR that would be useful for jjui developers.

flake.nix
doprz: Got it, partitions are new to me but I will implement this soon. Thank you for letting me know.

doprz:
> > I am glad this is getting rid of the `vendor-hash`, which was a minor pain for me after updating dependencies.
> 
> It is still [there](https://github.com/idursun/jjui/pull/426/changes#diff-206b9ce276ab5971a2489d75eb1b12999d4bf3843b7988cbe8d687cfde61dea0R41), not just in a file which was used to keep track of version changes on a single file. you'd still have to make vendorHash value empty and build to get the new vendored hash when adding dependencies.

I was thinking of using mod2nix but it seems a good part of the community has moved to just using buildGoModule. I believe some people said that it was using an old version of go but this could be resolved via an overlay. Open to suggestions.

.envrc
doprz: Great idea!

vic:
Using buildGoModule is fine for this repo, it is not a complicated Go compilation setup. Let's keep with standard buildGoModule and avoid adding more dependencies.

Our flake could, however, have a tiny package (a `pkgs.writeShellApplication` app) to help maintainers keep ./nix/vendor-hash updated. 

This is what I do

```
echo > nix/vendor-hash
nix build # this will fail since hash is empty and we need to take that hash to update vendor-hash
echo -n "the-hash" > nix/vendor-hash
```

we could have this script inside the development environment, and it could be run either manually (we could have a simple .bash script doing this) or we could have a GH action that runs ~on tags~ when go.sum changes and updates nix/vendor-hash.

doprz:
> Using buildGoModule is fine for this repo, it is not a complicated Go compilation setup. Let's keep with standard buildGoModule and avoid adding more dependencies.
> 
> Our flake could, however, have a tiny package (a `pkgs.writeShellApplication` app) to help maintainers keep ./nix/vendor-hash updated.
> 
> This is what I do
> 
> ```
> echo > nix/vendor-hash
> nix build # this will fail since hash is empty and we need to take that hash to update vendor-hash
> echo -n "the-hash" > nix/vendor-hash
> ```
> 
> we could have this script inside the development environment, and it could be run either manually (we could have a simple .bash script doing this) or we could have a GH action that runs ~on tags~ when go.sum changes and updates nix/vendor-hash.

Wouldn't having the `vendorHash` declared in the nix file instead of a text file be preferable?

vic:
> Wouldn't having the `vendorHash` declared in the nix file instead of a text file be preferable?

scripting to write into the plain-text file might be easier, or use `sed` to replace on the nix file. I've been using the plain-text file in other go repos of mine and for me it is better since the history of `go.sum` is reflected at `./nix/vendor-hash` without touching any nix specific logic. I'm good either way, I just have a preference for the plain text file and keep the history of go.sum and vendor-hash related one to another.

doprz:
> > Wouldn't having the `vendorHash` declared in the nix file instead of a text file be preferable?
> 
> scripting to write into the plain-text file might be easier, or use `sed` to replace on the nix file. I've been using the plain-text file in other go repos of mine and for me it is better since the history of `go.sum` is reflected at `./nix/vendor-hash` without touching any nix specific logic. I'm good either way, I just have a preference for the plain text file and keep the history of go.sum and vendor-hash related one to another.

Got it; that's a good point and it's nice to have `vendorHash` tracked with `go.sum`. Pushing changes now.

doprz:
Thank you for the detailed review @vic, I learned a lot from this! Addressed all PR comments + tested all features locally + edge cases, and now this PR is ready for a final review.

nix/jjui.nix
```diff
@@ -0,0 +1,78 @@
...
+    { pkgs, ... }:
+    let
+      buildInputs = [ pkgs.jujutsu ]; # Runtime
+      nativeBuildInputs = [ pkgs.makeWrapper ]; # Buildtime
+
+      jjui = pkgs.buildGoModule {
+        inherit buildInputs nativeBuildInputs;
+
+        pname = "jjui";
+        version = "dev";
```
vic: Looking at our current [history](https://github.com/idursun/jjui/commits/main/nix/default.nix) it looks like this PR is removing (perhaps unintended) versioning. We need [this](https://github.com/idursun/jjui/blob/main/nix/default.nix#L24) because people will be reporting issues and we might need them to provide the version the binary is built with.

doprz: I can do something like this: `gitCommit = inputs.self.rev or inputs.self.dirtyRev or "dev";` and pass that to `ldflags`.

nix/jjui.nix
```diff
@@ -0,0 +1,78 @@
+{ lib, ... }:
+{
+  perSystem =
+    { pkgs, ... }:
+    let
+      buildInputs = [ pkgs.jujutsu ]; # Runtime
```
vic: Hm, I don't think this is needed, because the postInstall step you have is **also** adding jujutsu to PATH.

nix/jjui.nix
```diff
@@ -0,0 +1,78 @@
...
+            ./../internal
+            ./../test
+          ];
+        };
+        vendorHash = lib.strings.trim (builtins.readFile ./vendor-hash);
+        doCheck = false;
+
+        postInstall = ''
+          wrapProgram $out/bin/jjui \
+            --prefix PATH : ${pkgs.lib.makeBinPath [ pkgs.jujutsu ]}
```
vic: I'm a bit worried about this, because this is is making the result binary use a fixed jujutsu version. The one from nixpkgs. Most of the time this will work, but many people (like me) would at times need to use more-recent versions of jujutsu than that currently at nixpkgs.

The way this is commonly solved is to have a nix file for the package and use pkgs.callPackage on it, so people is able to override the jujutsu package they want to use.

nix/jjui.nix
```diff
@@ -0,0 +1,78 @@
...
+          description = "A Text User Interface (TUI) designed for interacting with the Jujutsu version control system";
+          homepage = "https://github.com/idursun/jjui";
+          license = pkgs.lib.licenses.mit;
+          maintainers = [ "idursun" ];
+          platforms = pkgs.lib.platforms.unix;
+          mainProgram = "jjui";
+        };
+      };
+
+      update-vendor-hash = pkgs.writeShellApplication {
```
vic: lets move this helper app into its own file (another flake-parts module) since it is not related to packaging jjui itself.

nix/jjui.nix
doprz: The idea behind this is to have a pinned `jj` version to make sure `jjui` supports it. True, they could use an overlay but `pkgs.callPackage` seems like the defacto way.

flake.nix
```diff
@@ -1,9 +1,53 @@
...
+  outputs =
+    inputs@{
+      self,
+      flake-parts,
+      ...
+    }:
+    flake-parts.lib.mkFlake { inherit inputs; } {
```
vic: can you make this a oneliner that just imports a module from ./nix ? like:

`inputs: inputs.flake-parts.lib.mkFlake { inherit inputs; } ./nix` 

lets keep flake.nix as a manifest and have a main flake-parts entrypoint at ./nix.

flake.nix
```diff
@@ -1,9 +1,53 @@
...
+          {
+            imports = [
+              inputs.treefmt-nix.flakeModule
+              ./nix/shells.nix
+              ./nix/treefmt.nix
+            ];
+          };
+      };
+
+      flake = {
```
vic: the overlay output can also be its own module inside ./nix

flake.nix
```diff
@@ -1,9 +1,53 @@
...
+        ./nix/jjui.nix
+      ];
+
+      partitionedAttrs = {
+        devShells = "dev";
+        formatter = "dev";
+        checks = "dev";
+      };
+
+      partitions.dev = {
```
vic: partitions.dev, partitionedAttrs and the import of `flakeModules.partitions` all of them are semantically related make them a single module at ./nix

vic:
amazing work, added a few more comments. would you like to add a GH action that runs only when go.sum is modified and creates a PR with the updated ./nix/vendor-hash ? Because I believe idursun does not have Nix locally installed, and this would greatly reduce the amount of manual work.

doprz:
> amazing work, added a few more comments. would you like to add a GH action that runs only when go.sum is modified and creates a PR with the updated ./nix/vendor-hash ? Because I believe idursun does not have Nix locally installed, and this would greatly reduce the amount of manual work.

Thank you and will do!

nix/jjui.nix
vic: I think it is best we not to wrap jjui, I know it is a runtime dependency, wrapping would make any other jj on PATH just being ignored since the package will always use the one that comes with nixpkgs. What do you think? Nix people can easily wrap the result of this flake if they needed.

doprz: I got used to doing this for packages such as `lazygit` to ensure compatability but it seems that in jjui's docs:
Minimum supported `jj` version is `v0.26+` and this flake is using `v0.36.0` so it should work just fine. Agreed, Nix users can wrap it themselves, but in this case I would leave `jujutsu` in `buildInputs` so users have `jj` when using `jjui`. Thoughts?

doprz:
Also @idursun @vic should we add ourselves to this section as Nix maintainers?

https://github.com/idursun/jjui/blob/3fa9783a94dbf2f2bc7fe182a9335b6c4890bb05/README.md?plain=1#L116C1-L130C4

vic:
> https://github.com/idursun/jjui/blob/3fa9783a94dbf2f2bc7fe182a9335b6c4890bb05/README.md?plain=1#L116C1-L130C4

+1 on adding me and others that can help reviewing the Nix code.

EDIT: Oh, I'm not a nixpkgs maintainer, so I was not referring to adding me at the README as I do not maintain the nixpkgs jjui package. I was referring to [this](https://github.com/idursun/jjui/pull/426/changes#diff-6bd76ef7f80328de7fcf914f85a6a23f6719bf23b60467a73869a77699dd9c26R37) maintainers list. If I remember correctly that is a list of people keeping the nix code.

or perhaps a codeowners file at ./nix/ directory so we could be notified when people ask for reviews.

nix/jjui.nix
vic: ~Having only buildInputs is fine I guess.~ lets get rid of the wrappers.

Hm, buildInputs is for runtime-dependencies for installStep and friends? our current nix code has no buildInputs and everything has been fine. I'm now wondering if we even need buildInputs.

doprz:
> > https://github.com/idursun/jjui/blob/3fa9783a94dbf2f2bc7fe182a9335b6c4890bb05/README.md?plain=1#L116C1-L130C4
> 
> +1 on adding me and others that can help reviewing the Nix code.
> 
> EDIT: Oh, I'm not a nixpkgs maintainer, so I was not referring to adding me at the README as I do not maintain the nixpkgs jjui package. I was referring to [this](https://github.com/idursun/jjui/pull/426/changes#diff-6bd76ef7f80328de7fcf914f85a6a23f6719bf23b60467a73869a77699dd9c26R37) maintainers list. If I remember correctly that is a list of people keeping the nix code.
> 
> or perhaps a codeowners file at ./nix/ directory so we could be notified when people ask for reviews.

I'm also not on the nixpkgs maintainer's list so we can revisit this at a later time.

idursun:
> Because I believe idursun does not have Nix locally installed, and this would greatly reduce the amount of manual work.

I am using docker to get the vendor hash: 

```shell
docker run -it --rm -v "/Users/idursun/repositories/jjui:/project" -w /project nixos/nix:latest bash
```

I run the following inside bash. 

```shell
nix build --experimental-features "nix-command flakes" #.jjui
```

I guess this flow will still work.

idursun:
> perhaps a codeowners file at ./nix/

This is preferable because I believe GH will assign PRs directly to the ones in that list.

nix/shells.nix
```diff
@@ -0,0 +1,32 @@
...
+          gotools
+
+          jujutsu
+        ];
+
+        # Environment variables for development
+        CGO_ENABLED = "0";
+
+        shellHook = ''
+          # You can set JJUI_CONF_DIR to a gitignored directory inside your
```
vic: Having a shell hook is good idea, how about it just printing this message if JJUI_CONF_DIR environment is empty instead of people not knowing this is possible until they open the shell.nix file ?

doprz:
Ok @idursun @vic ready for a final final review! All steps have been re-tested, verified, and checked off in the first comment/description of this PR.

flake.nix
```diff
@@ -1,8 +1,14 @@
 {
+  description = "A Text User Interface (TUI) designed for interacting with the Jujutsu version control system";
+
   inputs = {
-    systems.url = "github:nix-systems/default";
-    nixpkgs.url = "github:nixos/nixpkgs/nixpkgs-unstable";
+    nixpkgs.url = "github:NixOS/nixpkgs/nixos-unstable";
     flake-parts.url = "github:hercules-ci/flake-parts";
+
+    flake-compat = {
```
vic: flake-compat can be removed from flake. it is pinned at default.nix / shell.nix.

doprz: Got it, it's because it uses `lock.nodes.root.inputs.flake-compat` which made it a required flake input.

nix/default.nix
```diff
@@ -1,42 +1,11 @@
...
-        nativeBuildInputs = [
-          pkgs.go
-          pkgs.gopls
-        ];
-      };
-    };
+  systems = inputs.nixpkgs.lib.systems.flakeExposed;
+
+  imports = [
+    inputs.flake-parts.flakeModules.partitions
```
vic: remove this import already imported inside partitions.nix

nix/package.nix
```diff
@@ -0,0 +1,45 @@
...
+    fileset = lib.fileset.unions [
+      ./../go.mod
+      ./../go.sum
+      ./../cmd
+      ./../internal
+      ./../test
+    ];
+  };
+  vendorHash = lib.strings.trim (builtins.readFile ./vendor-hash);
+  doCheck = false;
```
vic: any reason to don't run tests ? 

We [had](https://github.com/idursun/jjui/commit/16b3b6d7eb49fd6451f5b9d10abf812611086422) an explicit flake check, but seems it was removed by this PR.

vic (approved):
LGTM now, lets see if pipeline works and we see logs testing jjui on nix build.

@idursun could you please approve CI on this PR?

vic:
> a GH action that runs only when go.sum is modified and creates a PR with the updated ./nix/vendor-hash

@idursun would that be of value to you ? if so we maybe @doprz is interested in providing another PR after this one?

> I guess this flow will still work.

yes, same workflow would work with your dockered nix.

doprz:
Last thing. I used `mdformat` but I can also use `biome` or `prettier` to format markdown files. Any preference?

vic:
> Last thing. I used `mdformat` but I can also use `biome` or `prettier` to format markdown files. Any preference?

Hm, looks like the markdown formatter messed up the issues templates. Can we either skip those files via treefmt config or remove md formatting ?

.github/ISSUE_TEMPLATE/bug_report.md
```diff
@@ -1,19 +1,19 @@
----
+______________________________________________________________________
```
vic: revert these.

vic:
Don't know how to revert my approval at GH, haha.

doprz:
> Don't know how to revert my approval at GH, haha.

Make another review and select "request changes".

.github/ISSUE_TEMPLATE/bug_report.md
doprz: Will do. Interestingly `mdformat` messed this up.

vic (changes requested):
make sure md files frontmatter are not modified

.github/ISSUE_TEMPLATE/bug_report.md
doprz: `jj` makes this so easy.

doprz:
> make sure md files frontmatter are not modified

Done. I went with prettier and only for markdown files and it works as expected.

It also formatted `internal/config/default/config.toml` but it's still valid toml.

internal/config/default/config.toml
```diff
@@ -1,147 +1,147 @@
...
-    mode = ["o"]
-    restore = ["r"]
-    revert = ["R"]
-  [keys.file_search]
-    toggle = ["ctrl+t"]
-    up = ["up"]
-    down = ["down"]
-    accept = ["enter"]
-    edit = ["alt+e"]
+up = ["up", "k"]
```
vic: Hm, I know this is valid toml, but preserving spaces helps people maintaining the default config.

I'd prefer this not to be formatted. how about you @idursun.

doprz: I can exclude that file then to preserve the clarity.

vic:
I'm not really sure about all the formatting changes, they are not related to the title of this PR. Maybe we could work that on another PR if any. I'd prefer the toml files to be keep as they are, since having spaces aids people visually (and these toml files will also be read by users as reference so we need them to be as understandable and easy to read as possible). 

Can we keep this PR focused on the nix stuff ? 


EDIT: we'd also have to assume that Nix is just a convenience here, the main maintainer probably does not know how to run the nix formatters, if he agrees to keep formatters checks in place it'd be better to have a normal treefmt.yaml file and instructions on how to run treefmt outside the nix env. So we can have formatting not only on the nix environment.

doprz:
Ok. Dropping the formatting commit and commenting out the formatters. Other than that, I think we're done with this PR.

nix/treefmt.nix
```diff
@@ -0,0 +1,19 @@
+{
+  perSystem =
+    { pkgs, ... }:
+    {
+      treefmt = {
+        programs.nixfmt = {
+          enable = pkgs.lib.meta.availableOn pkgs.stdenv.buildPlatform pkgs.nixfmt-rfc-style.compiler;
```
Adda0, line 7: `nixfmt-rfc-style` is now just `nixfmt` and the previous name will be depecated, if I am not mistaken.

.github/CODEOWNERS
Adda0, line 1: @idursun Just to notify you as you should be aware of this. Depends on whether you want this behaviour or not.

.editorconfig
Adda0, line 1: @idursun Just to notify you. Depends on whether you want this or not.

Adda0 (commented):
While I am no expert, @vic most definitely is. ❤️  Amazing work. Not much more to say here. Mostly nits.

However, I feel like this PR should only touch the Nix stuff. Everything else (editorconfig, even nixfmt, formatting) should come with different PRs so @idursun can decide themselves whether they want the changes or not. If all the changes are desired, than I see no problem with the PR.

Highly educational Nix structure, too 👍

internal/config/default/config.toml
Adda0: Yeah, the same goes for me. And we should not force the formatting in a Nix stuff PR. If desired, another PR can be opened.

.editorconfig
idursun: Oh yes, I don't want this in this PR. Happy to review it in a different PR.

.github/CODEOWNERS
idursun: You mean the CODEOWNERS file? I agreed to have it in a different comment. PRs should suggest those names for review when changes are made to nix files.

internal/config/default/config.toml
idursun: I prefer nix steps to fail only if something related to nix flow is broken. 

I agree that the TOML needs formatting but that can be checked and verified separately. (e.g. Similar to `go fmt/go tidy` checks).

doprz: So this would also mean to disable `doCheck` in `nix/package.nix`? @vic Thoughts on this?
Ok, will remove formatting related things in this PR and include them in a new one.

.editorconfig
doprz: Ok I'll also include that as part of a new PR

internal/config/default/config.toml
vic: > So this would also mean to disable `doCheck` in `nix/package.nix`

No, checking that the nix package builds is part of what we already had. So don't disable doCheck, let the nix build run go tests, I'd even add a check that runs `jjui --version` just to make sure it builds and is runnable. In the first nix code I contributed this was not being checked, and then people improved this. So I think it is good for build-checks to stay.

nix/treefmt.nix
doprz: It seems the repo is `nixfmt` but the package name is still `nixfmt-rfc-style`
https://github.com/NixOS/nixfmt

doprz: But search.nixos.org shows `nixfmt` as the "Official formatter for Nix code" as of 25.11.

Ok I didn't get this warnings but the source code does verify this: https://github.com/numtide/treefmt-nix/blob/5b4ee75aeefd1e2d5a1cc43cf6ba65eba75e83e4/programs/nixfmt-rfc-style.nix#L20-L27

vic (approved):
LGTM again.

idursun:
Thanks @doprz @vic @Adda0 for all the work and the review!

doprz:
Thank you @vic @idursun @Adda0 for the reviews and help with this PR!

PR #427, opened 2025-12-15
feat(lua): add choose method and ui

idursun:
Added a Bubble Tea-based chooser overlay with optional title styling, and extended the Lua API with `choose`/`split_lines` so scripts can prompt for selections (positional options or table with options/title) and split text into lines.

The UI is barebones at the moment but gets the job done.

New Lua functions are:

  - choose(...options) → string|nil
  - choose({ options = { ... }, title = "Your title" }) → string|nil
  - split_lines(text: string, keep_empty?: boolean) → {strings}

```toml
[custom_commands."choose bookmark"]
  key_sequence = ["w", "i"]
  lua = '''
  local output, err = jj("bookmark", "list",  "-T", "name ++ '\n'")
  choice = choose({options = split_lines(output), title="choose a bookmark"})
  if choice then 
    flash("choice: " .. choice)
  else
    flash("none selected")
  end
  '''
```

image

### DIFF
--- a/internal/scripting/lua.go
+++ b/internal/scripting/lua.go
@@ -10,6 +10,7 @@ import (
 	"github.com/idursun/jjui/internal/ui/intents"
 	lua "github.com/yuin/gopher-lua"
 
+	"github.com/idursun/jjui/internal/ui/choose"
 	"github.com/idursun/jjui/internal/ui/common"
 	uicontext "github.com/idursun/jjui/internal/ui/context"
 	"github.com/idursun/jjui/internal/ui/operations/rebase"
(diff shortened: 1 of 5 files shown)


idursun:
@baggiiiie There's not a definite answer to that. It's usually a gut feeling after working on this project more than a year now.

I see core functionality at the level that we are interacting with the `jj` cli. (e.g. new commits, squash, rebase, etc)

The request you linked is a convenience feature on top of the core functionality. They are still very important but I would like them to be implemented through customisation. Customisation options we have might not be enough to implement those so then I prefer closing the gap in the customisation rather then implementing the feature in the core. There might be exceptions to this rule if implementing it in the core is more convenient, more performant or there's a huge ask about having it natively.

Additionally, if you can inflate a feature request in various ways then it is good sign that it should be a customisation. 

For example,
* I would like to copy the change id
* I would like to copy the commit id
* I would like to copy the commit id of the checked revisions
* I would like to copy the file name of the currently selected file
* I would like to copy the file name without the extension
* I would like to copy the file name as absolute/relative path
* I would like to copy the description
* I would like to copy the issue number from the description
* I would like to copy the file names that are checked

baggiiiie:
> if you can inflate a feature request in various ways then it is good sign that it should be a customisation

well put, totally agree!

just for discussion:
- in this very particular example of a yank panel, from my understanding of the codebase, i think certain info would be appropriate to implement natively (change/commit id, bookmark, file path in details view)
- for others you've listed (issue number from description, filename without extension) are highly individual dependent, and their implementations are doing more pollution the codebase rather than providing more convenient to users. in fact, for such requests, it might be *even easier* to implement with lua as complementary script than in source code

idursun:
> in this very particular example of a yank panel, from my understanding of the codebase, i think certain info would be appropriate to implement natively (change/commit id, bookmark, file path in details view)

Yes, totally. We should expose those through Lua. Except bookmark because we don't hold any information about the bookmarks. 

For the time being, users can get the commit id using something like `jj("log", "-r", revisions.current(), "-T", "commit_id", "--no-graph")`.

We need to implement something for exposing the current and checked files.

Issue #360, opened 2025-10-31
PageUp and PageDown always jump to the top / bottom of revision list


okuuva:
## Description
  
[This commit](https://github.com/idursun/jjui/commit/6e95d19846d33eb0fbd07d9563a2ad02cd9f9136) reintroduced a regression that was originally introduced in 0.9.3 and then fixed in 0.9.4. In the unaffected versions it was possible to bind `PageUp` and `PageDown` to e.g. scrolling the preview. After the linked commit, said buttons jump to the top/bottom of the revision list, making it impossible to use those keys for scrolling the preview. See #325, #327.

## Steps to Reproduce the Problem
A config snippet for ease of access:

```toml
[keys.preview]
scroll_up = ["pgup"]
scroll_down = ["pgdown"]
```

1. Bind `PageUp` and `PageDown` to scrolling the preview
1. Try scrolling the preview

## Expected Behavior
Preview scrolls up/down

## Actual Behavior
Revision cursor jumps to the top/bottom.

## Specifications

- Platform: macOS 26.0.1 (25A362)
- Version: 0.9.5

idursun:
Hey @okuuva thanks for the report and sorry for breaking your configuration. Currently, key bindings in revisions view take precedence over the others. I might expose the pgup/pgdown keybindings of the revisions view in the configuration so that you can unbind them if you want to assign them to the preview window. Let me have a look.

PR #429, opened 2025-12-16
oplog,list: refactor scrolling with Scrollable/StreamableList interface

baggiiiie:
### oplog,list: refactor scrolling with Scrollable/StreamableList interface
Refactored revisions/operation_log scrolling to have an IScrollableList
interface, with `scroll` method for navigation, which handles:
- Boundary detection (top/bottom of list)
- Page vs. single-step navigation
- Streaming/infinite scroll support
- User feedback messages (reaching top/bottom of list)

Also added a IStreamableList to handle streaming data that needs to by
loaded dynamically

### DIFF
--- a/internal/ui/common/list/list.go
+++ b/internal/ui/common/list/list.go
@@ -1,6 +1,10 @@
 package list
 
-import "io"
+import (
+	"io"
+
+	"github.com/idursun/jjui/internal/ui/common"
+)
 
 type IList interface {
 	Len() int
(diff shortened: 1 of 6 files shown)


oliverpool:
I can confirm that this fixes #360 with the following config: :+1: 
```toml
[keys]
scroll_down = []
scroll_up = []

[keys.preview]
half_page_down = ["pgdown"]
half_page_up = ["pgup"]
```

internal/ui/common/list/list.go
```diff
@@ -16,3 +20,92 @@ type IItemRenderer interface {
 	Render(w io.Writer, width int)
 	Height() int
 }
+
+// IListScroll defines the interface for list models that support navigation
+type IListScroll interface {
```
idursun: I am not sure about this interface, more specifically the shape of it. At first glance, it has `IList` and `IListCursor` which means this is a list and it will have a cursor. But then there's navigationconfig, which has ContextName, NavigationMessage, some concept of streaming all munched together. If I were to implement something that implements this interface, then I would have to look at its implementations first to make sense what all those mean. (Obviously I know what they mean, but I am saying all these things from the point of someone who is interacting with the codebase the very first time).

I would probably call it `IScrollableList` which conveys the message that this is a list, and it is scrollable. And then I would add methods that would make sense for a scrollable list (e.g. `VisibleRange() (int, int)` 

I would probably move the streaming related methods to a new interface (e.g. `IStreamableList`). 

These are just ideas that I came up with in a few mins, the actual implementation can be different but probably will be along those lines.

internal/ui/common/list/list_test.go
```diff
@@ -0,0 +1,336 @@
...
+			ContextName:   "test list",
+		},
+	}
+
+	result := Scroll(mock, 1, false)
+
+	if result.NewCursor != 0 {
+		t.Errorf("Expected cursor to remain 0 for empty list, got %d", result.NewCursor)
+	}
+	if result.NavigateMessage != nil {
```
idursun: You should be able to use `assert.NotNil` here.

internal/ui/oplog/operation_log.go
```diff
@@ -59,6 +81,8 @@ func (m *Model) ShortHelp() []key.Binding {
 	return []key.Binding{
 		m.keymap.Up,
 		m.keymap.Down,
+		m.keymap.ScrollUp,
+		m.keymap.ScrollDown,
```
idursun, line 83: These scroll up/down clean up is good. I am happy to merge those if you can extract those into a separate PR.

baggiiiie: drafted another PR #437, have a look!

internal/ui/common/list/list.go
baggiiiie: thanks for the feedback!
yup makes sense! the previous interface wasn't clear and a `NavigationConfig` is quite a vague concept to capture all information needed for scrolling.

i have updated `IScrollableList` and `IStreamableList`, much cleaner now

internal/ui/common/list/list_test.go
baggiiiie: done!

Issue #428, opened 2025-12-16
Can't exit if Git lock issue on startup?


shyun3:
## Description

I started `jjui` one day and this error showed up in the lower right hand pop-up window:
```
Error: Failed to reset Git HEAD state
Caused by:
1: Could not acquire lock for index file
2: The lock for resource '/path/to/.git/index' could not be obtained immediately after 1 attempt(s). The lockfile at '/path/to/.git/index' might need manual deletion.
```
(Exact path to lock file has been redacted.)

There was no content in the main windows. The log window just said `(no matching revisions)`.

I'm not sure why this happened, but that's not the focus of this ticket. The problem is I couldn't exit `jjui` afterward. I tried pressing `q`, `Esc`, and `Ctrl+C`. None of them succeeded in quitting or closing the error window. In the end, I had to press `Ctrl+Z` to suspend the process and manually kill it.

Even if some unrecoverable error occurred, I feel that it should still be possible to easily exit.

## Specifications

- Platform: WSL Ubuntu 24.04.3 LTS
- Version: 0.9.8

baggiiiie:
> The problem is I couldn't exit jjui afterward. I tried pressing q, Esc, and Ctrl+C. None of them succeeded in quitting

gosh that's horrible
does `jj` run? this error is probably from jj itself

baggiiiie:
i tried to reproduce the error, with the same error i have no issue quitting `jjui:0.9.7/0.9.8` with `q` tho

im on MacOS, doesn't have a WSL Ubuntu env to reproduce

for anyone wanting to check, `touch .git/index.lock` first and run any jj-git interop commands will trigger the error (e.g., jj describe)


<img width="2032" height="1167" alt="Image" src="https://github.com/user-attachments/assets/8563fe5e-4df7-4885-b8aa-5c28d1e7393b" />

shyun3:
I tried those steps, but they only seem to emit the error when running the command (e.g. `jj describe`). For my case, the error occurred immediately when starting `jjui`. I'm not exactly sure how to reproduce this, sorry if that's not very helpful.

baggiiiie:
> sorry if that's not very helpful.

no worries at all! hope someone with a similar setup could help reproduce this issue

shyun3:
Reporting that this happened to me again on 0.9.10. I guess another thing worth pointing out is that I'm often launching `jjui` from within Neovim using [vim-floaterm](https://github.com/voldikss/vim-floaterm).

[...]

PR #432, opened 2025-12-16
flash: add maxWidth to flash msg rendering

baggiiiie:
when i was looking at #428, i found out that if content for flash messages is too long, it could actually go beyond current window's width, instead of being line-wrapped

hence, im adding a maxWidth (currently set to 1/2 of screen) to the flash messages with line wrapped

## before
image

## after

image

### DIFF
--- a/internal/ui/flash/flash.go
+++ b/internal/ui/flash/flash.go
@@ -98,13 +98,32 @@ func (m *Model) View() []FlashMessageView {
 
 	y := m.Height - 1
 	var messageBoxes []FlashMessageView
+	// reserve padding and calculate max width for messages
+	maxWidth := m.Width - 4
+
 	for _, message := range messages {
-		var content string
+		var text string
+		var style lipgloss.Style
 		if message.error != nil {
-			content = m.errorStyle.Render(message.error.Error())
+			text = message.error.Error()
+			style = m.errorStyle
 		} else {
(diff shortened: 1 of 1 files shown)


internal/ui/flash/flash.go
```diff
@@ -97,13 +97,19 @@ func (m *Model) View() []FlashMessageView {
 	}
 
 	y := m.Height - 1
+	// set max width to half of the screen
+	maxWidth := m.Width / 2
+
```
idursun: What happens if we wrap by the whole width? I think this will wrap everything at the half width mark which is not what we want. (especially when the terminal window is split 1/2 or 1/3, which I do sometimes)

baggiiiie: > What happens if we wrap by the whole width? I think this will wrap everything at the half width mark which is not what we want. (especially when the terminal window is split 1/2 or 1/3, which I do sometimes)

good point!

baggiiiie:
i realize there's something off with the previous implementation, below line is getting all characters in messages, rather than getting the max width of lines in the message. 
https://github.com/idursun/jjui/blob/3a4ae1eb36b4a3d7ab55ae1432ee722bc7c1ea46/internal/ui/flash/flash.go#L108

added a check for natural rendering of flash message content, and only set width when it overflows

PR #434, opened 2025-12-18
feat: add .editorconfig

doprz:
https://editorconfig.org/

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,14 @@
+root = true
+
+[*]
+end_of_line = lf
+insert_final_newline = true
+charset = utf-8
+
+[*.go]
+indent_size = 2
+indent_style = tab
+
+[*.nix]
+indent_size = 2
+indent_style = space


.editorconfig
```diff
@@ -0,0 +1,8 @@
+root = true
+
+[*]
+end_of_line = lf
+insert_final_newline = true
+charset = utf-8
+indent_style = space
+indent_size = 2
```
idursun: Can we get rid of these as go recommendation is to use tabs instead of spaces

.editorconfig
```diff
@@ -0,0 +1,16 @@
...
+[*]
+end_of_line = lf
+insert_final_newline = true
+charset = utf-8
+indent_size = 2
+
+[*.go]
+indent_style = tab
+
+[Makefile]
```
vic: I get the feeling like if I was vibecoding haha, AIs always add extra things that are not needed. We don't have Makefile in the repo, this can be removed imho.

doprz: Working on a PR to add a `Makefile` right now. 3 PRs in parallel. No vibecoding here, just pure vim.

doprz: ~I can keep this in if you'd like if not then I"ll remove it (or include in the Makefile PR)~
Adding this in the Makefile PR for consistency.

.editorconfig
```diff
@@ -0,0 +1,13 @@
+root = true
+
+[*]
+end_of_line = lf
+insert_final_newline = true
+charset = utf-8
+indent_size = 2
```
idursun: Sorry, I guess I was not so clear with my previous comment. `indent_style` and `indent_size` should be all moved to their specific sections. I just don't prefer editor config to enforce any style globally, but it is fine if is done in a section.

In this case, `indent_size` can be moved to `nix` section. 

We can drop the `go` section as I am pretty sure `gofmt` will ignore and apply the defaults anyway.

doprz: Thanks for clarifying. I'll make those changes in a new commit.

However, I'd like to keep the `[*.go]` section. While gofmt does handle formatting, the `.editorconfig` ensures that editors apply the correct indentation while developers are writing code, _before_ gofmt runs.

This means:
- Code looks correct as you type, not just after formatting
- Developers can jump in without needing to adjust their personal editor settings

For example, use a no-config neovim instance to edit code and you will see that it will indent with spaces at the default width (8 spaces), which then needs to be corrected by gofmt. With `.editorconfig`, neovim uses tabs from the start, matching Go/jjui's conventions immediately.

doprz: Thoughts @vic + others?

idursun: Ah, that makes sense. Ok let's go with this. We can always revisit it.

PR #442, opened 2025-12-20
Add input and log to custom_commands API

ArnaudBger:
```
[custom_commands.test_input_and_log]
  key = ["T"]
  lua = '''
  local user_input = input({
    title = "Test Input and Log",
    prompt = "Enter a message: "
  })
  if user_input and user_input ~= "" then
    log("You entered: " .. user_input)
    flash("Logged: " .. user_input)
  else
    log("No input provided")
    flash("No input provided")
  end
  '''
```

### DIFF
--- a/internal/scripting/lua.go
+++ b/internal/scripting/lua.go
@@ -7,6 +7,7 @@ import (
 
 	"github.com/atotto/clipboard"
 	tea "github.com/charmbracelet/bubbletea"
+	"github.com/idursun/jjui/internal/ui/input"
 	"github.com/idursun/jjui/internal/ui/intents"
 	lua "github.com/yuin/gopher-lua"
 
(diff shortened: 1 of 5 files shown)


internal/scripting/lua.go
```diff
@@ -251,6 +255,18 @@ func registerAPI(L *lua.LState, runner *Runner) {
 		msg := L.CheckString(1)
 		return yieldStep(L, step{cmd: intents.Invoke(intents.AddMessage{Text: msg})})
 	})
+	logFn := L.NewFunction(func(L *lua.LState) int {
+		msg := L.CheckString(1)
+		logDir := fmt.Sprintf("%s/.config/jjui/custom_commands/%s", os.Getenv("HOME"), runner.commandName)
```
vic: Not sure that ~/.config is the proper place to write logs.

See [XDG dirs](https://wiki.archlinux.org/title/XDG_Base_Directory#User_directories).

ArnaudBger: Updated!

.gitignore
```diff
@@ -38,3 +38,5 @@ result-*
 
 # direnv
 .direnv
+
+.opencode-start.sh
```
idursun: Is this intended? I don't think this belongs to this repo.

internal/scripting/lua.go
```diff
@@ -139,7 +143,7 @@ func (r *Runner) Done() bool {
 	return r.done && r.await == nil
 }
 
-func registerAPI(L *lua.LState, runner *Runner) {
+func registerAPI(L *lua.LState, runner *Runner, commandName string) {
```
idursun: I understand commandName is passed to this because of the log function but I'd recommend it to keep it out of this PR. Custom commands are going to removed in the next release and I don't want to tie Lua script to custom commands by doing this.

.gitignore
ArnaudBger: Removed !

internal/scripting/lua.go
ArnaudBger: I did remove the log part !

PR #450, opened 2025-12-31
Search in preview panel

abourget:
I left a few comments with questions in the PR review page, and opened a discussion here: https://github.com/idursun/jjui/discussions/449

The feature works relatively well, and I think I'll keep it around for my personal use if it doesn't merge, but I'd much rather have something meaningful for lots of people.


internal/ui/preview/preview.go
```diff
@@ -198,14 +205,36 @@ func (m *Model) Update(msg tea.Msg) tea.Cmd {
...
+	m.updateViewportContent()
+}
+
+func (m *Model) updateViewportContent() {
+	if m.searchQuery == "" {
+		m.view.SetContent(m.content)
+		return
+	}
+
+	// Highlight matches with underline-only ANSI codes
```
abourget, line 232: I tried the configurable lipgloss we have in the toml files, but it issued an ANSI reset, and I use some colored diffs, so it broke the highlighting of the underlying diff tool.
underline and underline reset doesn't affect the diff's output as much.. which is why I landed on this.

internal/ui/ui.go
```diff
@@ -120,6 +121,16 @@ func (m *Model) handleFocusInputMessage(msg tea.Msg) (tea.Cmd, bool) {
...
+		// Handle preview keys when in details mode
+		if _, ok := m.revisions.CurrentOperation().(*details.Operation); ok {
+			switch msg.String() {
+			case "/":
+				return m.status.StartPreviewSearch(), true
+			case "n", "N", "g", "G":
+				return preview.PreviewCmd(msg), true
+			}
+		}
+
```
abourget, line 133: maybe the search actually belongs to the `preview` widget and not the details one.. but details view is the one that is altering the set of bindings, detaching from the revisions view.. so I'm unsure where to put it.

idursun:
closing this as there are so many conflicts

PR #451, opened 2025-12-31
Save YOffset between files, keep a global YOffset position _per file_ for the preview panel.

abourget:
Before this, the YOffset of the preview pane would jump, because it would keep the memory of the previous offset, whatever the file was there previously.
Now if you're inspecting a patch at a given offset, it'll keep it.

A prior implementation used the change_id in the _key_, to keep persistence across commits, might be a good idea to bring back.

Using a global, it allows to keep the memory even if you go back and forth other revisions. Would leak minimal memory here, perhaps could implement some sort of TTL with that memory.

---

Whoops, this brings another PR I had done. Please review only the last commit here!

### DIFF
--- a/internal/ui/preview/preview.go
+++ b/internal/ui/preview/preview.go
@@ -24,6 +24,8 @@ const (
 
 var _ common.Model = (*Model)(nil)
 
+var globalFileYOffsets = make(map[string]int)
+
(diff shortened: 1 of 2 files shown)


idursun:
closing due to conflicts

PR #452, opened 2026-01-03
feat: add $width variable for preview commands

pablospe:
## Summary
- Add a new `$width` placeholder variable that exposes the actual view width (in columns) to preview commands
- Enables tools like delta to use `--side-by-side` with the correct width
- Width updates dynamically when the preview pane is resized

## Motivation

When using delta with `--side-by-side` in preview commands, there's no way to pass the correct width because jjui pipes command output to an internal buffer (no TTY). This is similar to [Discussion #163](https://github.com/idursun/jjui/discussions/163).

Similar to how fzf exposes `$FZF_PREVIEW_COLUMNS`, this adds a `$width` variable.

## Example usage

```toml
[preview]
revision_command = ["util", "exec", "--", "bash", "-c",
  """jj show --color always --git -r $change_id | delta --side-by-side --width $width"""]
```

### DIFF
--- a/internal/jj/commands.go
+++ b/internal/jj/commands.go
@@ -9,12 +9,12 @@ import (
 )
 
 const (
-	ChangeIdPlaceholder    = "$change_id"
-	CommitIdPlaceholder    = "$commit_id"
-	FilePlaceholder        = "$file"
-	OperationIdPlaceholder = "$operation_id"
-	RevsetPlaceholder      = "$revset"
-	WidthPlaceholder       = "$width"
+	ChangeIdPlaceholder     = "$change_id"
+	CommitIdPlaceholder     = "$commit_id"
+	FilePlaceholder         = "$file"
+	OperationIdPlaceholder  = "$operation_id"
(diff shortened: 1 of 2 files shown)


idursun:
Hey thanks,

I’d suggest renaming the placeholder to `$preview_width` for specificity as it is only populated by the preview. 

Also, `go fmt` is needed.

pablospe:
> I’d suggest renaming the placeholder to $preview_width for specificity as it is only populated by the preview.

The initial implementation used `$preview_width`, but I renamed it to `$width` because the upcoming `feature/configurable-diff-command` branch (PR #453) will reuse this placeholder in diff commands as well. A generic `$width` avoids the situation of having `$preview_width` in non-preview contexts. But I can change it if you prefer, let me know.

> Also, go fmt is needed.

done, thanks!

pablospe:
> I’d suggest renaming the placeholder to `$preview_width` for specificity as it is only populated by the preview.

done! This PR is ready, let me know otherwise.

Issue #366, opened 2025-11-12
Search and filetree in diffs


abourget:
## Is your feature request related to a problem? Please describe.

I need to be able to search diffs, and jump to files in a diff, to study it fast

I use `gitk` usually for this. So only missing things here:
- being able to jump to a file when viewing a diff,
- being able to search the diff, to find the chunk of code I want to understand better

## Describe the solution you'd like

I think hitting `d` to view a diff is great, but then, in there, I'd like to have 4 more shortcuts:
- `/` to be able to search inside the diff view.. just plain text.. and go there
- `N` and `n` to go back and forth the matches, like in `vim` (why not `g` and `G`, end/home, like vim)
- `t` that would pop up a fuzzy search, to get down to the file I want.. and also see quickly which files _are_ int he diff when the patches are long. 
    - I could already jump to files that I know the name with the search.. but seeing the list is often useful, to see the sort of broadness of patches, the impact of it too.

## Describe alternatives you've considered

Right now, to study the diffs, I need to have this command:

`diffnav` has a great `t` shortcut, with fuzzy search. See the demo here: https://github.com/dlvhdr/diffnav
It looks like GitHub's `t`.. It's missing the `/` search though, making diffnav not ideal either, and it brings you outside, in a less familiar place.

Here's my command to get to it:
```
[custom_commands]
"diffnav from checked" = { args = ["util", "exec", "--", "bash", "-c", """
jj diff --from $checked_commit_ids --to $change_id --color=always --summary --git | diffnav
"""], show = "interactive" }
```

(allows me to diff between separate commits by choosing one.. that I haven't found a way to do natively in `jjui` and I do that all the time in `gitk`).

baggiiiie:
very rough first thought on this:
- right now diff view is simply rendering the content of a diff
- hence it's losing the pager's built-in functionalities like `/`, `n/N`, `g/G`
- a possible solution is to spawn a new view and let user's pager (like `delta`) take over, jjui only intercepts certain keys like `esc` to quit
https://github.com/idursun/jjui/blob/1356d11a3fc193f9ed142028010cbef34b128ee3/internal/ui/diff/diff.go#L54-L65

vic:
Yes, this is something that I always wanted (tend to use vim keys in pager -- and they just did not work inside jjui diffview). 

I like the idea of not having our own pager, and let whatever the user has configured as jj pager.

[...]

Issue #314, opened 2025-09-24
Custom pager: `delta` with `side-by-side` diff window incorectly sized


IzawGithub:
## Description

Using jjui with the [delta pager](https://github.com/dandavison/delta) with delta `side-by-side` option enabled makes the diff window of `jjui` incorrectly sized.

## Steps to Reproduce the Problem

1. Install `delta` (`0.18.2`)
2. Configure it with:
`~/.gitconfig`
```gitconfig
[delta]
    side-by-side = true

# Add this if you want to check the expected output from a `git diff`
[core]
    pager = delta
```
OR
```sh
export DELTA_FEATURES=+side-by-side
```
3. Launch a repo with `jjui`
4. Open a diff view with `d`, or check a preview with `p`

## Expected Behavior

What delta should look like:

<img width="2554" height="851" alt="Normal delta" src="https://github.com/user-attachments/assets/512cfcb2-64e4-4297-ad21-00d721dc466f" />

## Actual Behavior

How delta actually looks like:

<img width="2554" height="1436" alt="jjui delta" src="https://github.com/user-attachments/assets/bc1744b2-eb76-462b-84c9-74b5037d69c9" />

## Specifications

- Platform: Tested on two machine:
  - Ubuntu 25.04 6.14.0-29-generic x86_64
  - Arch 6.18.8-zen2-1-zen x86_64
- Version: `0.9.3`

idursun:
Hey, thanks.

I think this was discussed in #54 but without a proper solution. I don't know exactly why this is happening but my guess is that `delta` is not getting the terminal size information because it is started within `jjui` and defaulting to a size of 50 or something like that.

Unfortunately, I don't currently have the bandwidth to investigate this, but I will get back to this at some point.

krksgbr:
I also ran into this. Here's how I fixed it:

Since recently, jj passes a `$width` parameter to external diff tools: https://github.com/jj-vcs/jj/pull/7502.
It hasn't been released yet, so I did: `cargo install --git https://github.com/jj-vcs/jj jj-cli`

**~/.config/jj/config.toml**:
```toml
[ui]
diff-formatter = ["delta", "--width", "$width", "$left", "$right"]
```

TeddyHuang-00:
> Since recently, jj passes a `$width` parameter to external diff tools

Shortly after your comment, jj ships this feature into the release of 0.34.0, and I gladly took the change to try your solution.

[...]

PR #453, opened 2026-01-03
feat: add configurable diff command with $width and interactive support

pablospe:
## Summary
- Add a new `[diff]` config section with customizable `command` for the built-in diff view (d key on files)
- Support `show = "interactive"` to open diff in a real terminal (like custom commands)
- Pass screen width via `$width` placeholder for proper side-by-side diff rendering

## Motivation

When using delta with `--side-by-side` in the built-in diff view, there was no way to:
1. Pass the correct terminal width (delta would default to half screen)
2. Open an interactive terminal (required for proper pager behavior)

This complements the `$width` variable added for preview commands (#452).

## Example configuration

```toml
[diff]
# Use delta with side-by-side for the built-in diff view (d key on files in details view).
# +g: start at top of file (not bottom for small files)
command = ["util", "exec", "--", "bash", "-c", """jj diff --color always --git -r $change_id $file | delta --side-by-side --width $width --pager 'less -R +g'"""]
show = "interactive"

[custom_commands."deltadiff"]
# Custom command for viewing full revision diff with delta.
# This is needed because [diff] only applies to individual files (d key in file list),
# while this custom command applies to entire revisions (d key on revision list).
key = ["d"]
args = ["util", "exec", "--", "bash", "-c", """jj show --summary --git --color always -r $change_id | delta --pager 'less -R +g'"""]
show = "interactive"
```

## Available placeholders
- `$change_id` - revision change ID
- `$commit_id` - revision commit ID  
- `$file` - selected file path
- `$width` - screen width in columns

### DIFF
--- a/internal/config/config.go
+++ b/internal/config/config.go
@@ -169,7 +169,8 @@ func GetPreviewPosition(c *Config) (PreviewPosition, error) {
 }
 
 type DiffConfig struct {
-	Command []string `toml:"command"`
+	Command []string   `toml:"command"`
+	Show    ShowOption `toml:"show"`
 }
 
 type OpLogConfig struct {
--- a/internal/ui/operations/details/details.go
+++ b/internal/ui/operations/details/details.go
@@ -131,13 +131,16 @@ func (s *Operation) internalUpdate(msg tea.Msg) tea.Cmd {
 			if selected == nil {
 				return nil
 			}
+			args := jj.TemplatedArgs(config.Current.Diff.Command, map[string]string{
+				jj.ChangeIdPlaceholder: s.revision.GetChangeId(),
+				jj.CommitIdPlaceholder: s.revision.CommitId,
+				jj.FilePlaceholder:     selected.fileName,
+				jj.WidthPlaceholder:    strconv.Itoa(s.context.ScreenWidth),
+			})
+			if config.Current.Diff.Show == config.ShowOptionInteractive {
+				return s.context.RunInteractiveCommand(args, common.Refresh)
+			}
 			return func() tea.Msg {
-				args := jj.TemplatedArgs(config.Current.Diff.Command, map[string]string{
-					jj.ChangeIdPlaceholder: s.revision.GetChangeId(),
-					jj.CommitIdPlaceholder: s.revision.CommitId,
-					jj.FilePlaceholder:     selected.fileName,
-					jj.WidthPlaceholder:    strconv.Itoa(s.context.ScreenWidth),
-				})
 				output, _ := s.context.RunCommandImmediate(args)
 				return common.ShowDiffMsg(output)
 			}


pablospe:
**Note:** This PR is stacked on top of #452 (adds `$width` variable for preview commands). Please merge #452 first.

baggiiiie:
fyi the issue with delta side-by-side has a workaround as discussed in https://github.com/idursun/jjui/issues/314#issuecomment-3367379291

linting is failing, please run go fmt

pablospe:
Thank! Let me explain a bit better what is the problem. When pressing `d` on a file in `jjui`, the built-in diff view runs the diff command inside the TUI. With delta's `--side-by-side`, two issues occur:

1. **Width detection fails** - Delta can't detect the terminal width when running inside jjui's TUI, so side-by-side columns render incorrectly
2. **Pager doesn't work** - Delta's pager (less) needs an interactive terminal, which isn't available in the embedded view

**The workaround in #314** avoids the problem by disabling side-by-side mode entirely (`side-by-side = false`). That works, but you lose the side-by-side feature.

**This PR** solves it properly by:
- Adding `$width` so delta knows the exact terminal width
- Adding `show = "interactive"` to open the diff in a real terminal where the pager works

This lets you use delta with `--side-by-side` as intended, without compromises.

idursun:
Hey thanks,

This side by side diff issue when using delta is coming up regularly and at this point I am just ignoring it. I don't feel comfortable adding a configuration block just to enable integration one of the tools out there. 

My take on this is that;

If you want to use a custom pager for diffs, you can unbind the `d` from the internal diff view and define a custom command that binds to `d` and invokes the external diff viewer in an interactive session. You already have the command for that. I understand `delta` doesn't read the width of the terminal and defaults to 50 and it is the root cause here. I would instead have a look at why that's the case, because it should be able to query the terminal dimensions itself. Maybe our interactive session is missing some env variables for it to be able to do that or maybe that feature doesn't exists at all. In either case, I don't think the solution is to customise `jjui` to patch that gap.

baggiiiie:
> avoids the problem by disabling side-by-side mode entirely (side-by-side = false). That works, but you lose the side-by-side feature.

hey @pablospe , i'm using delta side-by-side with no issue at all, using the config suggested in #314 ,

[.gitconfig](https://github.com/baggiiiie/dotfiles/blob/691075b6dbc53ed7fbe457f672bd0a28fbd4d935/.gitconfig#L86-L95)
```
[delta "split-view"]
    side-by-side = true

[delta "non-split-view"]
    side-by-side = false

[delta]
    features = split-view
    navigate = true
    line-numbers = true
```
[jj/config.toml](https://github.com/baggiiiie/dotfiles/blob/691075b6dbc53ed7fbe457f672bd0a28fbd4d935/.config/jj/config.toml#L26-L34):
```
[ui]
diff-formatter = [
    "bash",
    "-c",
    "delta --width $width $left $right --features=split-view || true",
]
pager = "delta"
```

image

pablospe:
@baggiiiie Thanks! One question. Is this working for you when you enter to delta from `jjui` (when opened from the file list `l` shortcut)?
`navigate = true  # use n and N to move between diff sections`

baggiiiie:
> Is this working for you when you enter to delta from jjui (when opened from the file list l shortcut)?

display works fine when showing diff from file list, tho it's only display, losing pager's other functionalities (regarding pager, related discussion see #366 )

[asciicast](https://asciinema.org/a/765919)

pablospe:
> losing pager's other functionalities

Yes, exactly, thanks for the video and the confirmation. I thought I was missing something. This is exactly what I noticed and wanted to solve with this PR, which allow me to have the full delta experience with n/N navigation (pager functionalities).

pablospe:
> you can unbind the d from the internal diff view and define a custom command that binds to d and invokes the external diff viewer in an interactive session

@idursun 
Before we abandon this PR. How would you do that? I mean, unbind and bind again? I tried with this configuration:
```toml
  [keys]
  diff = []

  [keys.details]
  diff = []

  [custom_commands.deltadiff]
  key = ["d"]
  args = ["util", "exec", "--", "bash", "-c", "jj diff --git --color always -r $change_id $file | delta --pager 'less -R +g'"]
  show = "interactive"
```
but it doesn't seem to be working:
  - `d` is only working in the revision list
  - `d` is *not* working In the details view (after pressing l) → how do I re-bind `d` to shows diff for a specific file with a custom command?

pablospe:
@idursun ignore my previous message. I found a solution for what I wanted to achieve with this PR, I think the inline format made the difference (somehow).

In jjui config file:
```toml
# Unbind built-in diff key to allow custom command to handle it
[keys]
diff = []

# Interactive diff with delta (works in both revision list and details view)
# - $change_id: replaced with selected revision
# - $file: replaced with selected file (in details view) or empty (in revision view)
# - show = "interactive": opens real terminal so delta's pager/navigation works
[custom_commands]
"deltadiff" = { key = ["d"], args = ["util", "exec", "--", "bash", "-c", "jj diff --git --color always -r $change_id $file | delta --pager 'less -R +g'"], show = "interactive" }
```

@baggiiiie Could you give a try to this? It will give you the pager's functionalities. And you don't need to touch .gitconfig and jj/config anymore, only jjui/config.

baggiiiie:
@pablospe nice thanks, your custom command works!

pablospe:
Ok, we can close this PR. Thanks for the help both and the very useful tool, @idursun and @baggiiiie.

Issue #433, opened 2025-12-17
IP: `context` methods in Lua


idursun:
We have `revisions.current()` in Lua bridge that returns the change id of the currently selected revisions. We are not on par with supporting all the placeholders that are available in the legacy custom commands.

I would like to make them available through `context` variable.

* `jjui.context.change_id()`
* `jjui.context.commit_id()`
* `jjui.context.file()`
* `jjui.context.checked_files()`
* `jjui.context.checked_change_ids()`
* `jjui.context.checked_commit_ids()`
* `jjui.context.operation_id()`

I like this API more than providing access to them through the UI element name(e.g. `revisions.current()`) If we are to follow the same convention then `file` should be placed under `details.file()` which might suggest that it is always available even if the `details` is not showing.

vic:
+1 on having them unified under `context`, makes sense. will absence (eg, no checked file selected) be signaled via `nil` or empty list?

idursun:
> will absence (eg, no checked file selected) be signaled via nil or empty list?

I was thinking of `nil` because it makes `if` conditions easier. I also don't have any experience with Lua so I am open to recommendations!

baggiiiie:
when i was first playing around with our lua support when it's introduced, i subconsciously wrote something like:
```lua
  local rev = revisions.current()
  local commit_id = rev.commit_id()
  copy_to_clipboard(commit_id)
```

i have nothing against the `context` as proposed.

though i find it more intuitive to write `revision.change_id()`, as that's how my mental model is, after getting familiarity with `jj`'s template language

when it comes to implementation, would it also be easier to return a "revision" object with lua API, parsed from existing Commit struct? 
https://github.com/idursun/jjui/blob/b1a2a8007ad4d75c0f6f76300c958f772d5c20f1/internal/jj/commit.go#L9-L14

idursun:
> when it comes to implementation, would it also be easier to return a "revision" object with lua API, parsed from existing Commit struct?

It would definitely be easier but not necessarily better. It would ultimately couple the Lua bridge with the jjui internals, which might change in the future.

PR #454, opened 2026-01-05
Add yank to clipboard feature for file paths

remo5000:
Allows you to copy files to paste/use in other apps. 

https://github.com/zellij-org/zellij 's mouse selection doesn't work that well with jjui, likely from how the TUI library works, but tapping 'y' is easier than clicking

### DIFF
--- a/internal/config/default/config.toml
+++ b/internal/config/default/config.toml
@@ -75,6 +75,7 @@ limit = 0
     diff = ["d"]
     select = ["m", " "]
     revisions_changing_file = ["*"]
+    yank = ["y"]
   [keys.evolog]
     mode = ["v"]
     diff = ["d"]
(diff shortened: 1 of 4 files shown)


baggiiiie:
nice thanks for contributing

i personally like this features and have a local build that supports it. 
though i have a different approach to it: it's implemented with lua api, details see https://github.com/idursun/jjui/commit/7d92982eb77ac1d53813cee526f4f37d92b111a6 

and with this build i use this config to support copying either ChangeID or filename, depending on if details view is open:
```toml
[custom_commands.copy_to_clipboard]
key = ["Y"]
lua = '''
  if details.is_open() then
    local file = details.current_file()
    copy_to_clipboard(file)
    flash("Copied to clipboard: " .. file)
  else
    local rev = revisions.current()
    copy_to_clipboard(rev)
    flash("Copied to clipboard: " .. (rev or "none"))
  end
  '''
```

i think @idursun might like lua script better too, would like to hear what you think

idursun:
Hey thanks, 

Currently `y` is already bound to `duplicate` so this new keybinding won't work. Additionally, there has been many discussions about similar feature asks in the past and I still believe the best way forward is to enable these features through customisations.

As @baggiiiie mentioned, this is better suited for a custom command which the current lua scripting is capable of copying contents to clipboard. However, current lua script bindings lack the capability of getting the list of the checked file names. There's an open issue about this here #433 Once that's implemented and merged, this should be possible.

remo5000:
@idursun that makes sense to me, will close this out! FWIW, the bindings don't clash since we only listen for 'y' in the details view. @baggiiiie thanks for landing `details` into lua, very nice!

baggiiiie:
hey @remo5000, #433 is implemented and now i have this custom commands to yank change_id or file path:

```toml
[custom_commands.copy_to_clipboard]
key = ["Y"]
lua = '''
local selected_file = context.file()
if selected_file then
    -- A file is selected, which means we're in details view
    -- Copy the file path
    copy_to_clipboard(selected_file)
    flash("Copied file: " .. selected_file)
else
    -- No file selected, copy the change ID
    if change_id then
        copy_to_clipboard(change_id)
        flash("Copied change ID: " .. change_id)
    else
        flash("No item selected to copy")
    end
end
'''
```

PR #456, opened 2026-01-06
feat: make flash message display time configurable

### LINKED ISSUE #455: FR: Allow configuring flash message expiration time
Currently, informational flash messages are displayed for a hard-coded four seconds.  This frequently isn't enough time for me to read the message and/or to copy information out of it if desired.

I propose making the flash message timeout configurable (though continuing to default to 4 seconds).  I have developed a PoC on my machine which uses the config key `ui.flash_message_display_seconds`.

Other ideas I've considered would be implementing a keystroke which would halt the flash message expiration timer, leaving the message up until manually dismissed with the ESC key.  However, this would still require that action be taken within the 4-second window.  Another possibility would be to implement the ability to restore the most recent flash message.  I'm not sure if either of these would be better from a UX point of view compared to simply making the timeout configurable, but they would definitely be more complex to implement.  I have also considered (but not implemented in my PoC) making a special value for `ui.flash_message_display_seconds` (possibly `0` or `-1`) meaning "never expire".  However, I don't know if jjui uses sentinels like that for other configuration settings or not (the same effect could be achieved with a very large timeout).

AFAIK, with the current implementation informational flash messages are displayed for four seconds and then are deleted with no way to see them again and no way to make them display for longer.

## Additional context
I'm happy to submit a PR if this feature seems agreeable.

living180:
Replace the hard-coded display time of four seconds for informational flash messages with a configurable display time.  The config key is `ui.flash_message_display_seconds`, defaulting to `4` for consistency with the previous behavior.  The special value of `0` means display messages until manually dismissed.

fixes #455

### DIFF
--- a/internal/config/config.go
+++ b/internal/config/config.go
@@ -8,6 +8,7 @@ import (
 	"os/exec"
 	"path"
 	"strings"
+	"time"
 )
 
 //go:embed default/*.toml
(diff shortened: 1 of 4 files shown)


idursun (approved):
Thanks! Looks good to me.

Issue #309, opened 2025-09-20
FR: Interactive bookmarks list view


OliverJAsh:
## Is your feature request related to a problem? Please describe.
I have many bookmarks representing WIP which I frequently need to switch between.

Currently I have to either find the bookmark in the revlog or press `:` then type `edit <bookmark>` or `new <bookmark>` (no autocomplete).

## Describe the solution you'd like
A bookmarks list view that allows me to select a bookmark and perform actions on it (e.g. `edit` or `new`).

## Describe alternatives you've considered
None

## Why current features fail to cover this request
As per above, the UX could be much better.

## Additional context
lazygit shows a list of branches and you can press space to checkout the branch, or `n` to create a new branch from there.

lazyjj has a bookmarks list view with these shortcuts to interact with the selected bookmark in the list:

```
Main panel
j/k     scroll down/up
J/K     scroll down by ½ page
a       show all remotes
c       create bookmark
r       rename bookmark
d/f     delete/forget bookmark
t/T     track/untrack bookmark
Enter   view in log
n       new from bookmark
N       new and describe
e       edit bookmark
```

idursun:
Hey, thanks!

> Currently I have to either find the bookmark in the revlog or press : then type edit <bookmark> or new <bookmark> (no autocomplete).

You can press `B` to open bookmark editor on the selected revision. The editor will suggest the the movable bookmark names so you can partially type and then hit `tab` to to autocomplete and hit `enter` to effectively move the bookmark to the selected revision (or create a new bookmark if it doesn't exist)

There's currently no support for renaming bookmarks, but that's only because I haven't worked on an intuitive way to be able to do something like that.

For an interactive bookmark list, `b` should open the bookmark menu. It contains pretty much every bookmark related operation (expect rename). I usually do `bmm` to move main to the selected revision and `bt` and `bu` to track/untrack etc. 

Not sure if you haven't seen it or if it didn't do what you wanted to do.

OliverJAsh:
Thanks for the quick reply!

It seems the behaviour you're describing is for managing bookmarks, but I think I'm asking for something different. That is, the ability to edit or "checkout" a bookmark, similar to running `jj bookmark list`, finding the bookmark name, and then running `jj edit <bookmark>` or `jj new <bookmark>`.

[...]

PR #463, opened 2026-01-09
fix(bookmarks): track new bookmarks

idursun:
After the merge of #407 I noticed that we need to track local bookmark before we can push them. 

The bookmark menu doesn't show an option to track a local bookmark because it doesn't have any remotes. This change adds the option to track local bookmarks. 

The only problem is track command tracks a bookmark for all the remotes if it the remote is not specified. Currently, we cannot do this as the bookmark menu doesn't allow choosing a remote but I am hoping to address this when the bookmark list is remodelled around the discussion in #309

### DIFF
--- a/internal/jj/bookmark_parser.go
+++ b/internal/jj/bookmark_parser.go
@@ -28,6 +28,10 @@ func (b Bookmark) IsDeletable() bool {
 	return b.Local != nil
 }
 
+func (b Bookmark) IsTrackable() bool {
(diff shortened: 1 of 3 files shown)


OliverJAsh:
I'm not sure if this is working, or am I doing something wrong?

https://github.com/user-attachments/assets/1bb4aa2d-1d5b-4a23-a218-1209b2115eee

idursun:
@OliverJAsh You need to track it from `bookmarks` window first and then you can `git push` it.

PR #468, opened 2026-01-14
feat: add "no edit" option when splitting to skip editing messages

thomaschrstnsn:
I would like to avoid editing descriptions whenever I am splitting commits.

This is my take at solving this. 

Apologies for skipping issue creation as per contribution guidelines, this was more natural for me for this small change.

### CHANGES

- Add a `split.no_edit` configuration option.
- Prevent editor from opening when splitting commits.
- Set default `no_edit` value to `false`.

### DIFF
--- a/internal/config/config.go
+++ b/internal/config/config.go
@@ -26,6 +26,11 @@ type Config struct {
 	Limit     int               `toml:"limit"`
 	Git       GitConfig         `toml:"git"`
 	Ssh       SshConfig         `toml:"ssh"`
+	Split     SplitConfig       `toml:"split"`
+}
+
+type SplitConfig struct {
+	NoEdit bool `toml:"no_edit"`
 }
 
 type Color struct {
--- a/internal/config/default/config.toml
+++ b/internal/config/default/config.toml
@@ -152,3 +152,6 @@ limit = 0
 
 [ssh]
   hijack_askpass = false
+
+[split]
+  no_edit = false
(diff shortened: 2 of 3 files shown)


Adda0 (commented):
suggestion(non-blocking): While I would absolutely love to be able to configure this, too, is this not more of an upstream jj concern? You can set, as you do here, `--message = ""`, but there is no option in the jj configuration to set this globally for all split (and other) operations. I suggest opening an upstream jj issue to discuss configuring this and whether and for which operations it makes sense.

nickchomey:
I'd also love to see this functionality, one way or another.

idursun:
Hey, thanks but you can already do this today without having to change any code.

All you need to do to unbind the default `s` and then bind it to the custom command you want:

```toml
[keys]
"split" = [""] # this will unbind the default key

[custom_commands]
"split no edit" = { key = ["s"], args = ["split", "-r", "$change_id", "-m", ""], show = "interactive" }
```

As per the PR, I don't want and like introducing configuration options for supporting personal customisations. They are hard to maintain and very like to break as I won't be using any of them.

Currently, we have custom commands to support this use case today and I am also working on new ways to make this more powerful.

thomaschrstnsn:
Nice, that makes sense. I also had a hunch that it could be achieved with binds but could not figure it out. 

Thanks for taking the time to teach me, and thanks for a great TUI! 🙌

PR #471, opened 2026-01-15
lua: expose exec_shell to lua script

baggiiiie:
this is to allow commands like below, which opens a file for quick edits from jjui:
```toml
[custom_commands.open_file]
key = ["O"]
lua = '''
local file = context.file()
if not file then
    flash("No file selected")
    return
end
exec_shell("vim " .. file)
'''
```

ps: i tried using `os.execute` but it wouldn't return to jjui after vim exits. hence exec_shell function is exposed via lua.


idursun (approved):
Thanks!

PR #472, opened 2026-01-15
ui,preview: fix preview pane ctrl-u/d scrolling

baggiiiie:
Commit 8481b92 broke ctrl-u/d scrolling in the preview pane, see below
code:
```
if common.IsInputMessage(msg) {
    if m.oplog != nil {
        cmds = append(cmds, m.oplog.Update(msg))
    } else {
        cmds = append(cmds, m.revisions.Update(msg))
    }
    return tea.Batch(cmds...)
}
```
This causes unhandled key messages to be routed only to oplog or
revisions, and then returns early, so they never reach the preview
model's Update() call which happens later in the function.

This fix groups preview related commands and make sure they are always
handled.

Also, added a YOffset() method to internal/ui/preview/preview.go to expose
the viewport's Y offset for testing purposes.

### DIFF
--- a/internal/ui/preview/preview.go
+++ b/internal/ui/preview/preview.go
@@ -110,6 +110,10 @@ func (m *Model) WindowPercentage() float64 {
 	return m.previewWindowPercentage
 }
 
+func (m *Model) YOffset() int {
+	return m.view.YOffset
+}
+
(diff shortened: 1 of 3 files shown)


idursun (approved):
Thanks!

I was almost sure that change was going to break something but couldn't find anything myself.

PR #475, opened 2026-01-18
fix: set description from stdin to allow messages starting with minus

### LINKED ISSUE #315: Inline describe fails with messages starting with hyphen(`-`)
## Description

When entering a message starting with hyphen(`-`) in *Inline describe*, an error is shown.

1. Select commit
1. press `Enter` key
1. Enter `-foo`
1. press `Ctrl-s`

## Expected Behavior

Commit messages will be set to `-foo`.

## Actual Behavior

The following error message is displayed:
```
error: unexpected argument '-f' found

  tip: to pass '-f' as a value, use '-- -f'

Usage: jj describe [OPTIONS] [REVSETS]...

For more information, try '--help'.
```

## Specifications

- Platform: Windows 11 Pro 24H2
- Version: jjui `v0.9.4-0.20250924201740-9721a260ebcf`, jj `0.33.0`

Edit:
Originally I wrote `--`, but since it also happens with a single hyphen, I corrected it.

kankri:
Setting the description from stdin avoids command line parsing issues when the message cannot be properly quoted.

E.g. `jj describe -m -foo` gives the error `error: unexpected argument '-f' found`.

This fixes #315.

### DIFF
--- a/internal/jj/commands.go
+++ b/internal/jj/commands.go
@@ -30,6 +30,10 @@ const (
 )
 
 type CommandArgs []string
+type CommandWithStdin struct {
+	Args  CommandArgs
+	Input string
+}
 
 func ConfigListAll() CommandArgs {
 	return []string{"config", "list", "--color", "never", "--include-defaults", "--ignore-working-copy"}
@@ -97,8 +101,11 @@ func Describe(revisions SelectedRevisions) CommandArgs {
 	return args
 }
 
-func SetDescription(revision string, description string) CommandArgs {
-	return []string{"describe", "-r", revision, "-m", description}
+func SetDescription(revision string, description string) CommandWithStdin {
+	return CommandWithStdin{
+		Args:  []string{"describe", "-r", revision, "--stdin"},
+		Input: description,
+	}
 }
 
 func GetDescription(revision string) CommandArgs {
--- a/internal/ui/context/command_runner.go
+++ b/internal/ui/context/command_runner.go
@@ -22,6 +22,7 @@ type CommandRunner interface {
 	RunCommandImmediate(args []string) ([]byte, error)
 	RunCommandStreaming(ctx context.Context, args []string) (*StreamingCommand, error)
 	RunCommand(args []string, continuations ...tea.Cmd) tea.Cmd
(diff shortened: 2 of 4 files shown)


internal/jj/commands.go
```diff
@@ -30,6 +30,10 @@ const (
 )
 
 type CommandArgs []string
+type CommandWithInput struct {
+	Args  CommandArgs
+	Input string
+}
```
baggiiiie: if this struct is only associated with `--stdin`, maybe `CommandWithStdin` is clearer

```suggestion
type CommandWithStdin struct {
	Args  CommandArgs
	Stdin string
}
```

baggiiiie (commented):
hey @kankri thanks for the PR

since this is an issue with jj (as pointed out in [this comment](https://github.com/idursun/jjui/issues/315#issuecomment-3337186512)), it might be better to wait for jj to handle `jj describe -m "--test"`

on the other hand, do you know if there's a way to actually pass in `-` as the first character to `jj describe`?

kankri:
> since this is an issue with jj (as pointed out in [this comment](https://github.com/idursun/jjui/issues/315#issuecomment-3337186512)), it might be better to wait for jj to handle `jj describe -m "--test"`

Commit messages can be long and contain arbitrary characters, I'm afraid to find out later what other corner cases there are in addition to the most visible issue with "clap" parsing and the initial minus character.

Even if Jj eventually fixes passing minus as the 1st character in the "-m" option, I feel that using "--stdin" is much safer, especially in Windows where the command line is a single string with a limited length and needs to be constructed by quoting with some [arcane rules](https://learn.microsoft.com/en-us/windows/win32/api/shellapi/nf-shellapi-commandlinetoargvw#:~:text=CommandLineToArgvW%20has%20a%20special%20interpretation%20of%20backslash%20characters).

internal/jj/commands.go
kankri: It's not associated just with `--stdin` option, but could be used in other commands which can take input from the standard input which is given as a string. Though now that I'm checking, "describe" seems to be currently the only one which reads stdin.

In any case, `CommandWithStdin` is probably the better name. I guess I was thinking there could be also a case for running a command with stdin from an iterator rather than from a string and didn't want to use a longer name like `CommandWithStdinFromString`...

PR #477, opened 2026-01-19
preview: reset preview offset after revision change

baggiiiie:
Previously, if a revision change occurred while a preview was open, the
preview offset stays the same, which causes new preview to be shown at
an unexpected position. This commit resets the preview offset to 0.

see below for bug details:
jjui-preview-scrolling-offset


idursun:
I guess this is similar in spirit what #451 was trying to fix.

idursun (approved):
One line change. I am happy with that :)

PR #478, opened 2026-01-19
fix preview split border and stacked menu indexes

baggiiiie:
split border has a negative z index, which causes it to be hidden behind other UI elements.

This commit sets the z index to 1 to ensure the border is visible.

bug detail:
image

### DIFF
--- a/internal/ui/bookmarks/bookmarks.go
+++ b/internal/ui/bookmarks/bookmarks.go
@@ -305,7 +305,7 @@ func (m *Model) ViewRect(dl *render.DisplayContext, box layout.Box) {
 	frame := cellbuf.Rect(sx, sy, menuWidth, menuHeight)
 	if len(m.menu.VisibleItems()) == 0 {
 		fillRect := frame.Inset(1)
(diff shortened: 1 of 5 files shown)


idursun:
Hey, thanks! I had a look at this yesterday but couldn't get it done because it is more than just changing the z index of the split. We need to adjust the z index of stacked views (e.g. bookmarks, git, etc). Currently, they are using `menu` and it is by default rendering at z=0, hence leading to this visual artifact.

image

I'd say we need to set the default Z order of the menu a little bit higher than the split's. Feel free to have a stab at it.

baggiiiie:
> We need to adjust the z index of stacked views (e.g. bookmarks, git, etc).

u have it! i just found out this bug too and im working on it 🤠

baggiiiie:
on my end, im seeing menus not rendering properly too. title blended into main menu:
image

nickchomey:
hah! i saw this but didnt want to bother you folks with a bug report. Glad to see that you already solved it

PR #480, opened 2026-01-19
refactor: intents

idursun:
This PR brings the intents to the rest of the codebase as a continuation of the work started in #415. The reason for this change is to lay the foundation for bringing the changes discussed in the #319 

This is mostly a structured refactoring, and should be a transparent change to the end user. If it's not then, it's a bug.


idursun:
Mostly FYI type of PR. Merging to prevent potential conflicts, I will follow it up with any changes necessary.

PR #487, opened 2026-01-23
fix: revset/ui, suggestion overlapping main view

baggiiiie:
When revset is triggered (with `L`), it pops open a suggestion view
below, which previously pushed `jj log` graph view down, but now overlaps.
It's not really a functionality issue, i've been thinking about if this
needs a "fix" for a few days and decided i prefer the previous aesthetic.

This commit adds `Height()` method to revset model, which returns 1 line
when not editing and actual height including suggestions when editing,
and it changes the overlay rectangle to start at box.R.Min.Y+1 

bug detail:
image

### DIFF
--- a/internal/ui/revset/revset.go
+++ b/internal/ui/revset/revset.go
@@ -190,6 +190,19 @@ func (m *Model) Update(msg tea.Msg) tea.Cmd {
 	return m.autoComplete.Update(msg)
 }
 
+func (m *Model) Height() int {
+	if !m.Editing {
+		return 1
+	}
+	// When editing, calculate height including suggestions
+	var w strings.Builder
+	w.WriteString(m.styles.promptStyle.PaddingRight(1).Render("revset:"))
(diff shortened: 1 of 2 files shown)


idursun:
Hey, thanks.

This was actually intentional. 

I had plans around showing additional information in the revset completion box than the list of functions. (e.g. history, bookmarks, tags, functions + signatures). I haven't get around to working on it though.

We can merge this but bear in mind it will get back to overlapping version to include those features. 

I am curious to hear what really bothered you about the current floating aesthetics (I admit it looks like a bug for folks who are used to the previous version)

baggiiiie:
@idursun it's definitely because i'm used to how it used to look and the current overlapping UI feels off

if i have to reason about it, i think it's the lack of border? i imagine if more content were to be fit in, a border would be nice and consistent with current ui layout

> We can merge this but bear in mind it will get back to overlapping version to include those features.

if this is intentional feel free to close this PR, or merge it if u see fit!

PR #490, opened 2026-01-25
feat: add target picker to some operations (e.g. rebase)

idursun:
This PR adds a convenience target picker which can be opened by pressing `t` in various operations (rebase, duplicate, revert, squash) which allows you to select a target bookmark/tag or free form text input to set the destination for the command. Pressing `alt+enter` adds `--ignore-immutable` argument to the command that ran.

Addresses some of the features discussed in #309 

https://github.com/user-attachments/assets/353013ea-0f4c-4586-8ad9-4a78b6bcf8ac


nickchomey:
This is a nice addition! 

Though I think it is going in a very different direction from what most were in support of in #309. 

My understanding was that it would be great to be able to pop up a list of all bookmarks and act upon them from there. Eg open the list, navigate to/fuzzy search for feature-1, then press r (which might automatically select B?) then navigate/search for main and press enter. 

Likewise, you could use the bookmark modal to do things like jump to the root or head of the bookmark, git operations etc. Having this modal tied to just selecting a target for rebase (or, presumably later, squash etc) seems both limiting and cumbersome - you'd be continually opening and closing it rather than just operating within it, like a jj log, but for bookmarks. 

A bookmarks control panel modal would be significantly better when there are many branches spread very far apart. Maybe you have no idea where feature-1 even is in order to navigate to it to begin with. But, I recognize that that might be a significant amount of work. 

So, if this target picker really is the direction jjui is going, perhaps it could instead be used moreso just for "jumping" - open it, find the branch (eg feature-1) to jump to, from there you can do whatever single operations you need to (edit description, git operations, move to another commit on the branch and move the bookmark), or you could press rebase and then open the jumper again, search for main, then press enter to complete the rebase. 

This might be the simplest way to make this very featurefull, as it just uses normal jjui for everything as opposed to making a separate command center for bookmarks. 

I hope this helps

idursun:
Hey @nickchomey 

You are right, bookmark view will be something different than this. This is something I came up with while working/iterating on a new bookmark view. I kept it around and found it useful over time.

The demo is only showing the use of a bookmark, but it also works with tags and free form revsets. You can type the target whatever you want it to be (e.g. `main- | xyz`). Additionally, it allows you to choose bookmarks that you are not tracking, as they don't show up in the revision list. 

In short, it is adding flexibility to what we already have.

internal/jj/commands.go
```diff
@@ -251,6 +251,11 @@ func BookmarkListAll() CommandArgs {
 	return []string{"bookmark", "list", "-a", "--template", allBookmarkTemplate, "--color", "never", "--ignore-working-copy"}
 }
 
+func TagList() CommandArgs {
+	const template = `name ++ "\t" ++ normal_target.commit_id().shortest(1) ++ "\n"`
```
baggiiiie: nit: seems like `name` is used in `targetArg()`, commit_id might not be necessary

idursun: thanks! fixed.

idursun:
> tho wondering if we should reuse what we have for `git` and `bookmark` for consistency
I actually removed the `menu` package and inlined all usages of it to `git` and `bookmark` models. Going forward, I believe each model should own its View implementation and rely on common view components as little as possible. Often, I find them similar yet subtly different. For instance, the remote filter in Git and bookmarks look the same but have subtly different functionalities. Git’s remotes define the scope for preloaded commands while bookmarks act as a command filter.

PR #493, opened 2026-01-27
refactor: centralize z-index constants in render/zindex.go

baggiiiie:
Replace scattered magic numbers with centralized `render.Z*` constants:
- ZBase (0): base content (revisions, diff, oplog, details)
- ZFuzzyInput (1): fuzzy inputs, revset list
- ZRevsetOverlay (2): revset overlay
- ZPreview (10): preview/split views
- ZDialogs (50): modal dialogs (undo/redo, input fields) - above preview
- ZMenuBorder/ZMenuContent (100/101): menus
- ZOverlay (200): flash/sequence overlays
- ZHelpPage (250): help page
- ZPassword (300): password modal

Also, in order to deal with confirmation in different render (`details`
vs `undo/redo`), WithZIndex is added for caller-specified z-index.
- Undo/redo use WithZIndex(ZDialogs) to render above preview.
- Details confirmation uses default ZBase to stay below preview.

This also fixes an issue with `confirmation` menu overlapping on preview:
image

### DIFF
--- a/internal/ui/confirmation/confirmation.go
+++ b/internal/ui/confirmation/confirmation.go
@@ -44,6 +44,7 @@ type Model struct {
 	Styles      Styles
 	messages    []string
 	stylePrefix string
+	zIndex      int
 }
 
 func (m *Model) ShortHelp() []key.Binding {
@@ -92,6 +93,13 @@ func WithOption(label string, cmd tea.Cmd, keyBinding key.Binding) Option {
 func WithAltOption(label string, cmd tea.Cmd, altCmd tea.Cmd, keyBinding key.Binding) Option {
 	return func(m *Model) {
 		m.options = append(m.options, option{label, cmd, keyBinding, altCmd})
+	}
+}
+
+// WithZIndex sets the z-index for rendering
+func WithZIndex(z int) Option {
+	return func(m *Model) {
+		m.zIndex = z
 	}
 }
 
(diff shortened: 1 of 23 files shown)


internal/ui/confirmation/confirmation.go
```diff
@@ -183,7 +183,7 @@ func (m *Model) ViewRect(dl *render.DisplayContext, box layout.Box) {
 	sy := box.R.Min.Y
 
 	frame := cellbuf.Rect(sx, sy, bw, bh)
-	window := dl.Window(frame, 10)
+	window := dl.Window(frame, 0)
```
idursun: I think due to this change undo, redo renders like this: 

image

internal/ui/input/input.go
```diff
@@ -105,8 +105,8 @@ func (m *Model) ViewRect(dl *render.DisplayContext, box layout.Box) {
 	content := lipgloss.JoinVertical(0, rows...)
 	content = m.styles.border.Padding(0, 1).Render(content)
 	box = box.Center(lipgloss.Size(content))
-	window := dl.Window(box.R, 10)
-	window.AddDraw(box.R, content, 1)
+	window := dl.Window(box.R, render.ZPreview)
+	window.AddDraw(box.R, content, render.ZFuzzyInput)
```
idursun: Something similar is happening with input too: 

Image

baggiiiie:
@idursun nice thanks, good catch!

it's tricky one since both `undo/redo` and `details` use `confirmation`, and in `undo/redo`, the pop up should be above `preview` and for `details` it should be beneath

patch with one more z index value, now:

confirmation for undo, redo:
- image

`confirmation` in `details` and `input`:
- image

Issue #240, opened 2025-07-29
FR: Add command to generate and follow a new pull request link


theherk:
**Is your feature request related to a problem? Please describe.**
After creating a new bookmark and pushing to a remote, the next step is often to go and request the patch be pulled / merged.

**Describe the solution you'd like**
Using one key to generate and follow this link is very handy. Tools like [lazygit](https://github.com/jesseduffield/lazygit/blob/master/docs/Config.md#custom-pull-request-urls) and [neogit](https://github.com/NeogitOrg/neogit/blob/master/README.md?plain=1#L103-L108) make this quite simple.(those link are to the documentation on the topic)

**Describe alternatives you've considered**
I could use a script to do this, no problem. I could even run that as an alias or custom command. I suspect, there will be users over time that seek this feature.

**Why current features fail to cover this request**
I can't promise that they do, but I haven't found the option at least.

---

I'd be willing to implement such a feature maybe, only if this is desired by @idursun and in accordance with other design goals.

idursun:
Hey thanks for the suggestion @theherk 

The version in the `main` now has flash messages, which is basically showing output of the ran command as a flash message on the side. Link to the PR is here: #206 

This includes showing the `git push` command output too.

Currently, error messages are stick (needs to dismissed by pressing `esc`) and the success messages disappear after 4 seconds.

What I am not sure about is if the `git push` and `git fetch` output messages should also be sticky even though they are successful because they usually contain important information as you described.

Does making `git push` command output sticky cover your use case? (as it usually contains a link to the PR)? or do you have a more general use case in mind?

theherk:
I'm not sure if that link comes back in all git host cases, though it may, but that is up to the host. If it does, that still requires following the link via clicking if available in that terminal context. In addition, this link will only be available at initial branch push time, and one might wish to create the request after pushing a few more commits or just generally in the future.

But what you say would absolutely help in many cases.

idursun:
>  In addition, this link will only be available at initial branch push time, and one might wish to create the request after pushing a few more commits or just generally in the future.

True! 

Having support for generating a link to PR on demand doesn't hurt IMO. Feel free to give it a go. 👍🏿

vic:
> Does making `git push` command output sticky cover your use case?

Maybe not the right solution for this issue, but it would also be nice to have a way to show previous flash messages. Many vim distributions or other editors have a way to show "notifications" that were already closed by the popup timer. I could work on this if it makes sense to you.

[...]

PR #494, opened 2026-01-27
feat: add create pull request feature to git menu

dancorreia-swe:
This PR adds the ability to create pull requests directly from jjui’s git menu proposed in #240. When a bookmark has been pushed to a remote, users can open the PR/MR creation page in their default browser. Currently, this PR supports GitHub, GitLab, and Bitbucket as providers.

The default branch can be configured via `config.toml` and is set to `main` by default.

How it works
- Open the git menu with `g`
- Press `r` to filter PR-related items
- Select a bookmark that has been pushed to a remote to open the PR creation page

https://github.com/user-attachments/assets/e7b699fb-2ef2-46aa-9afc-661985dbdd5c

edit: added the issue where this feature was proposed

### DIFF
--- a/internal/browser/browser.go
+++ b/internal/browser/browser.go
@@ -1,0 +1,24 @@
+package browser
+
+import (
+	"os/exec"
+	"runtime"
+)
+
+func Open(url string) error {
+	var cmd *exec.Cmd
+
+	switch runtime.GOOS {
+	case "darwin":
+		cmd = exec.Command("open", url)
+	case "linux":
+		cmd = exec.Command("xdg-open", url)
+	case "windows":
+		cmd = exec.Command("cmd", "/c", "start", url)
+	default:
+		// Fall back to xdg-open for other Unix-like systems
+		cmd = exec.Command("xdg-open", url)
+	}
+
+	return cmd.Start()
+}
--- a/internal/config/config.go
+++ b/internal/config/config.go
@@ -281,7 +281,8 @@ func GetSuggestExecMode(c *Config) (SuggestMode, error) {
 }
 
 type GitConfig struct {
-	DefaultRemote string `toml:"default_remote"`
+	DefaultRemote       string `toml:"default_remote"`
+	DefaultTargetBranch string `toml:"default_target_branch"`
 }
 
 func GetGitDefaultRemote(c *Config) string {
@@ -290,6 +291,14 @@ func GetGitDefaultRemote(c *Config) string {
 		return "origin"
 	}
 	return remote
+}
+
+func GetGitDefaultTargetBranch(c *Config) string {
+	branch := c.Git.DefaultTargetBranch
+	if strings.TrimSpace(branch) == "" {
+		return "main"
+	}
+	return branch
 }
 
 type SshConfig struct {
--- a/internal/config/default/config.toml
+++ b/internal/config/default/config.toml
@@ -108,6 +108,7 @@ limit = 0
     mode = ["g"]
     push = ["p"]
     fetch = ["f"]
+    create_pr = ["r"]
   [keys.oplog]
     mode = ["o"]
     restore = ["r"]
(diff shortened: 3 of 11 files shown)


oliverpool:
I think some providers output a URL in the git push log.
This URL is usually to open a PR (or visit an open PR, or see a CI job).

I think it would be pretty handy if this log could be parsed and an unobstrusive prompt could ask if the user whishes to visit this URL.

I will have to check, but I think this works with many forges, like GitHub, Forgejo/Codeberg, Gitlab…

Edit: just saw that this was mentioned in #240

idursun:
Hey, thanks for the work!

This looks nice but as I have mentioned in similar issues, this kind of features falls into personal customisations umbrella for me. I personally just click on the url that's shown in the flash message to create the PR after pushing to remote. I believe this can be implemented a custom command with what we have already.

I want all similar issues to be handled by using customisation methods rather than being implemented in the code base. Rule of thumb for me is that `jjui` should only invoke `jj` command internally and nothing else.

I labelled the original issue as _v0.10_ because initially v0.10 was supposed to have a very flexible customisation system. We have delivered the Lua support before v0.10, so the label doesn't make sense anymore. I will change the label to something like _customisations_ to make it more clear.

dancorreia-swe:
Got it! Thank you all for the feedback - it does make sense. I'll port the solution to a Lua script as you suggested.

idursun:
For the sake of completeness, I asked AI to generate a custom command based on your patch and this is what it came up with. Seems to work fine:

```toml
[custom_commands."Open GitHub PR"]
  desc = "Open GitHub PR for bookmark"
  key_sequence = ["w", "r"]
  lua = """
  -- Get the selected revision's change id
  if not change_id then
    flash("No revision selected")
    return
  end

  -- Get bookmarks on this revision (local bookmarks only)
  local template = [[if(!remote, name ++ "\n")]]
  local output, err = jj({"bookmark", "list", "-r", change_id, "--template", template, "--color", "never"})
  if err then
    flash("Failed to get bookmarks: " .. err)
    return
  end

  local bookmarks = {}
  for _, line in ipairs(split_lines(output)) do
    if line ~= "" then
      table.insert(bookmarks, line)
    end
  end

  if #bookmarks == 0 then
    flash("No bookmarks on this revision")
    return
  end

  -- If multiple bookmarks, let user choose
  local bookmark
  if #bookmarks == 1 then
    bookmark = bookmarks[1]
  else
    bookmark = choose({title = "Select bookmark for PR", options = bookmarks})
    if not bookmark then
      return
    end
  end

  -- Get remote URL for origin
  local remote_output, remote_err = jj({"git", "remote", "list"})
  if remote_err then
    flash("Failed to get remotes: " .. remote_err)
    return
  end

  local remote_url = nil
  for _, line in ipairs(split_lines(remote_output)) do
    local name, url = line:match("^(%S+)%s+(.+)$")
    if name == "origin" and url then
      remote_url = url
      break
    end
  end

  if not remote_url then
    flash("No origin remote found")
    return
  end

  -- Parse GitHub URL (supports both SSH and HTTPS)
  local owner, repo

  -- Try SSH format: git@github.com:owner/repo.git
  local ssh_host, ssh_path = remote_url:match("^git@([^:]+):(.+)$")
  if ssh_host and ssh_host:find("github") then
    owner, repo = ssh_path:gsub("%.git$", ""):match("^([^/]+)/(.+)$")
  end

  -- Try HTTPS format: https://github.com/owner/repo.git
  if not owner then
    local https_host, https_path = remote_url:match("^https?://([^/]+)/(.+)$")
    if https_host and https_host:find("github") then
      owner, repo = https_path:gsub("%.git$", ""):match("^([^/]+)/(.+)$")
    end
  end

  if not owner or not repo then
    flash("Not a GitHub repository or unsupported URL format")
    return
  end

  -- Build the PR URL (comparing against main)
  local pr_url = string.format("https://github.com/%s/%s/pull/new/%s", owner, repo, bookmark)

  -- Open in browser
  exec_shell("open " .. pr_url)
  """
```

PR #500, opened 2026-02-02
feat: add interactive split for files in details view

baggiiiie:
- Add `interactive` to `Split` and `Restore` command builders
- Remove `SplitInteractive` and `RestoreInteractive` functions as they
  are now redundant
- Add `IsInteractive` field to `StartSplit` and `DetailsSplit` intents
  for future keybinding support

### DIFF
--- a/internal/jj/commands.go
+++ b/internal/jj/commands.go
@@ -82,10 +82,13 @@ func DiffEdit(changeId string) CommandArgs {
 	return []string{"diffedit", "-r", changeId}
 }
 
-func Split(revision string, files []string, parallel bool) CommandArgs {
+func Split(revision string, files []string, parallel bool, interactive bool) CommandArgs {
 	args := []string{"split", "-r", revision}
 	if parallel {
 		args = append(args, "--parallel")
+	}
+	if interactive {
+		args = append(args, "--interactive")
 	}
 	var escapedFiles []string
 	for _, file := range files {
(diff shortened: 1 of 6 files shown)


internal/jj/commands.go
```diff
@@ -95,6 +95,17 @@ func Split(revision string, files []string, parallel bool) CommandArgs {
 	return args
 }
 
+func SplitInteractive(revision string, file string, parallel bool) CommandArgs {
```
idursun: Cannot we reuse the existing `Split` instead. We can add `interactive` parameter similar to `parallel`

internal/ui/operations/details/details.go
```diff
@@ -204,6 +205,9 @@ func (s *Operation) handleIntent(intent intents.Intent) tea.Cmd {
 			confirmation.WithOption("Yes",
 				tea.Batch(s.context.RunInteractiveCommand(jj.Split(s.revision.GetChangeId(), selectedFiles, intent.IsParallel), common.Refresh), common.Close),
 				key.NewBinding(key.WithKeys("y"), key.WithHelp("y", "yes"))),
+			confirmation.WithOption("Interactive",
+				tea.Batch(s.context.RunInteractiveCommand(jj.SplitInteractive(s.revision.GetChangeId(), selected.fileName, intent.IsParallel), common.Refresh), common.Close),
```
idursun: Instead of `selected.fileName` this should work with `selectedFiles`.

baggiiiie:
hey @idursun thanks!

i followed the existing `RestoreInteractive` and added `SplitInteractive`, since this isn't desirable, i removed them all together and updated to use `interactive` as an input arg. updated the intent accordingly

note that right now `RestoreOperation` is still using the old `RestoreOperationMsg` style, i only added `interactive` to the intent struct, left the bigger refactor untouched

idursun:
Thanks, I'll have a look at the restore operation separately.

Issue #298, opened 2025-09-06
FR: Make it possible to quite jjui from any "view"


AngelEzquerra:
## Is your feature request related to a problem? Please describe.
Currently it is only possible to quite jjui (using the "q" keyboard shortcut) from the "top level view". If you are on the "diff" view, for example, you must first hit "ESC" to exit the diff window, and then you can use "q" to exit. This is a bit annoying because the actions you must take to exit the app depend on your current "location" in the app UI.

## Describe the solution you'd like
I'd like to be able to always use the same keyboard shortcut (e.g. "que") to exit the app, no matter where you were in the app UI.

## Describe alternatives you've considered
I guess you could kill the app using ctrl-c, but that does not seem like such a good idea. It would imply to always use ctrl-c to exit, even if you were at the "top view".

idursun:
This will be possible in the v0.10 series: https://github.com/idursun/jjui/discussions/319

You will be able to bind to any key in any view, (e.g. for diff) but can also be applied to oplog, details views.

```toml
    diff = [
        { on = "j", do = "diff.down" },
        { on = "k", do = "diff.up" },
        { on = "ctrl+d", do = "diff.halfpagedown" },
        { on = "ctrl+u", do = "diff.halfpageup" },
        { on = "f", do = "diff.pagedown" },
        { on = "b", do = "diff.pageup" },
        { on = "esc", do = "close diff" },
        { on = "q", do = "quit" },
    ]
```

J0J0:
> I guess you could kill the app using ctrl-c, but that does not seem like such a good idea. It would imply to always use ctrl-c to exit, even if you were at the "top view".

For me, `ctrl+c` does _not_ kill jjui at all, even in the main view. For what it's worth, i'm currently using the following (pretty Linux specific) workaround in my `config.toml`:
```
[custom_commands.CtrlC]
desc = "Quit jjui"
key = ["ctrl+c"]
lua = '''
  local fh = io.open("/proc/self/stat", "r")
  local pid = fh:read("*n")
  fh:close()
  os.execute("kill " .. pid)
'''
```
Is there a better way to do this? If not, i'd appreciate a lua command for quitting. :)

baggiiiie:
> Is there a better way to do this? 

@J0J0 didn't expect people would use lua script to kill jjui. maybe the best is to open a lua API to quit jjui so there's no need for `io.open("/proc/self/stat")` this kind of operation

J0J0:
[...]

PR #505, opened 2026-02-04
feat: enable quit keymap for different modes

baggiiiie:
This should addresses some (or all?) of the issues in #298

### DIFF
--- a/internal/ui/operations/evolog/evolog_operation.go
+++ b/internal/ui/operations/evolog/evolog_operation.go
@@ -85,6 +85,8 @@ func (o *Operation) HandleKey(msg tea.KeyMsg) tea.Cmd {
 	switch {
 	case key.Matches(msg, o.keyMap.Cancel):
 		return o.handleIntent(intents.Cancel{})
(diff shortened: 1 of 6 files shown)


internal/ui/operations/evolog/evolog_operation.go
```diff
@@ -117,7 +117,15 @@ func (o *Operation) ShortHelp() []key.Binding {
 	if o.mode == restoreMode {
 		return []key.Binding{o.keyMap.Cancel, o.keyMap.Apply}
 	}
-	return []key.Binding{o.keyMap.Up, o.keyMap.Down, o.keyMap.Cancel, o.keyMap.Evolog.Diff, o.keyMap.Evolog.Restore}
+	return []key.Binding{
+		o.keyMap.Up,
+		o.keyMap.Down,
+		o.keyMap.Cancel,
+		o.keyMap.Quit,
```
idursun, line 126: Quit is not handled here though.

baggiiiie: good catch! fixed!

PR #510, opened 2026-02-06
feat: multi line revset completion

idursun:
This PR changes the revset completion to be multi line and loads it with bookmarks + tags + aliases + functions + history.

Adds support for new functions introduced in the recent versions of jj (e.g. divergent, bisect, author_*, etc)

Also does a little bit of refactoring to reduce the duplication around loading bookmarks and tags in various places (e.g. set bookmark, target picker and now revset completion)

image


idursun:
> wondering why is it call bkmrk rather than just bookmark, since there's plenty of space

I will change that. 

> and is it possible to have some default suggestions before user starts typing?

It can, however I thought when the revset is empty it should only show the history items because it was the default behaviour of up and down keys. It shows empty at first because there's nothing in the history yet. I can change it to show as much as possible while giving priority to history items.

idursun:
I am going to merge these changes but feel free to comment on the PR if you have concerns, we can always follow it up with another PR

PR #512, opened 2026-02-08
fix(ui): prevent custom commands from shadowing operation keys

baggiiiie:
When in an operation mode (e.g., rebase), custom command key bindings
were intercepting keys meant for the operation. 

For example, with a custom command bound to 's' (like suggested here
https://github.com/idursun/jjui/pull/468#issuecomment-3765421729) would
fire instead of the normal behavior of rebasing the current branch.

Guard both single-key and multi-key sequence custom command handling
behind an InNormalMode() check so operation keys take priority.

### DIFF
--- a/internal/ui/revisions/revisions.go
+++ b/internal/ui/revisions/revisions.go
@@ -923,6 +923,11 @@ func (m *Model) highlightChanges() tea.Msg {
 	return nil
 }
 
+// SetRows is a test helper that sets rows and selects the given revision.
+func (m *Model) SetRows(rows []parser.Row, selectedRevision string) {
+	m.updateGraphRows(rows, selectedRevision)
+}
+
 func (m *Model) updateGraphRows(rows []parser.Row, selectedRevision string) {
 	if rows == nil {
 		rows = []parser.Row{}
--- a/internal/ui/ui.go
+++ b/internal/ui/ui.go
@@ -205,10 +205,14 @@ func (m *Model) Update(msg tea.Msg) tea.Cmd {
 		}
 		return nil
 	case tea.KeyMsg:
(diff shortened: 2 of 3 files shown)


baggiiiie:
hey @idursun 
im converting this PR to draft, this is a bigger issue than i thought.

with this change, it means NO custom commands would be allowed when it's NOT in default mode. this is certainly not ideal. 
however, a custom command like https://github.com/idursun/jjui/pull/468#issuecomment-3765421729 would not be possible, either, as it makes `s` in rebase mode unusable

the only way i could think of to deal with this, is to have some sort of allow list in custom commands?

idursun:
@baggiiiie 

Custom commands are meant to be used at the very top level (revisions) with non-conflicting keybindings, but some of the key handling leaked to operations and this bug has become the feature.

I would like to address the whole key customisation and custom command short comings in the next big refactoring instead of patching use cases one by one.

> the only way i could think of to deal with this, is to have some sort of allow list in custom commands?

Yes, so each key binding should map to an action and there needs to be a scope definition defining when they can be active.

baggiiiie:
> each key binding should map to an action and there needs to be a scope definition defining

@idursun i can add something like a `mode` key to toml config, and filter what keys to pass when a custom command is triggered. is that what you have in mind? 

> address the whole key customisation and custom command short comings in the next big refactoring

how would u like to refactors this? anything i could help with?

idursun:
> how would u like to refactors this? anything i could help with?

Thanks for asking! I think it is best to create a tracking issue for this so it gets some visibility.

I created it here: https://github.com/idursun/jjui/issues/515

baggiiiie:
closing this in favor of #515

PR #521, opened 2026-02-12
feat(oplog): add QuickSearch to oplog, share logic with revision search

baggiiiie:
Extract shared CircularSearch utility and Searchable/SearchableLine
interfaces to enable search reuse across views. Wire quick search into
the oplog view with n/N cycling, Esc/Enter to clear, and highlight
matched text segments. Refactor revisions search to use the same shared
utility.

This is useful as when I open oplog, i usually intend to revert/restore
certain operation, and typing `/` to quick search keywords like `squash`
has been tremendously useful as compared to visual scans with my eyes :)


baggiiiie:
@idursun i've refactored this PR to on top of the newly merged #533

Issue #515, opened 2026-02-09
IP: Unified Action & Binding system


idursun:
Currently, jjui relies on `tea.KeyMsg` handling scattered across individual models. Each model has its own key matching logic in `Update()`, which means:

- Adding or changing a keybinding requires modifying Go code
- Users cannot customise bindings without forking the project
- Key handling is duplicated and inconsistent across models
- There's no single place to see what keys do what (e.g. command palette)

Over time we've added several customisation mechanisms — `[custom_commands]` for user-defined shell/Lua commands, `[leader]` for multi-key sequences, and `[keys]` for basic rebinding. These work but they're separate systems with different config formats, different capabilities, and different limitations. This new system would unify all of them into a single `[[actions]]` + `[[bindings]]` model.

I've been introducing **intents** as a way to decouple input from behaviour. Instead of models reacting to raw key presses, they respond to typed intent messages like `Navigate{Delta: -1}` or `Apply{Force: true}`. This is already partially done — most models now handle intents alongside legacy `KeyMsg` matching.

The next step is to close the loop: introduce a **central dispatcher** that converts key presses into actions, resolves those actions into intents, and dispatches them to the appropriate model. This would make keybindings fully user-configurable via a TOML config file.

### Proposal

The current `[keys]` config section is a flat key-to-function map. I'd replace it with two separate concepts: **actions** (e.g. named operations like `"revisions.move_up"`, `"rebase.apply"`, `"rebase.set_target"`) and **bindings** (mappings from keys to actions, scoped to specific UI contexts). Since bindings can carry arguments, the same action can be bound to multiple keys with different behaviour. For example, the `rebase.apply` action takes a `force` argument — we could bind it to `enter` with `force = false` and to `alt+enter` with `force = true`.

Each binding would live in a scope, so the same key can do different things depending on the state of the app. For example, j would move down in the revision list when in the revisions scope, but move the selected file down when the revisions.details scope is active. A central dispatcher would resolve key presses against the current scope and fire the matching action.

The bulk of the migration work is getting models to stop handling `tea.KeyMsg` directly and only respond to intents. The `tea.KeyMsg` case in `Update()` methods should be removed everywhere except true text-input widgets.

[...]

PR #527, opened 2026-02-13
Add customizable keybindings for the diff view ([keys.diff_view])

Dima-369:
Apologies for not opening an issue first to discuss this as per the contributing guidelines — the change felt small enough to submit directly.

The `[keys.diff_view]` config section is missing, although other modes like `[keys.preview]` and `[keys.details]` have one. I assume this was just an oversight. This PR adds it, making the diff view's scrolling, paging, and close keybindings customizable — the diff is relatively small since it follows the existing pattern.

One behavioral change: the separate `Cancel` (`esc`) and `Quit` (`q`) handlers are unified into a single `close` key (defaults to `["esc", "q"]`), which now always closes the diff view rather than `q` quitting the entire app. Happy to revert this part if you'd prefer keeping the quit behavior.

### DIFF
--- a/internal/ui/diff/diff_test.go
+++ b/internal/ui/diff/diff_test.go
@@ -31,7 +31,7 @@ func TestScroll_AdjustsViewportOffset(t *testing.T) {
 
 func TestUpdate_CancelReturnsClose(t *testing.T) {
 	model := New("content")
-	model.keymap.Cancel = key.NewBinding(key.WithKeys("q"))
+	model.keymap.DiffView.Close = key.NewBinding(key.WithKeys("q"))
 
 	var msgs []tea.Msg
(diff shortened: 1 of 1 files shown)


idursun:
Hey, thanks!

This makes sense. I can back port this to #515 as it is not handled there too.

I would like to keep the current behaviour (i.e. `q` quits), everything else seems fine to me.

Dima-369:
> I would like to keep the current behaviour (i.e. q quits), everything else seems fine to me.

Sure, can you check my last commit if that is the expected way to handle this?

idursun:
Hey, thanks! Overall looks good but there are test failures need fixing.

Dima-369:
Oops, forgot about them. Fixed now.

PR #530, opened 2026-02-14
Auto-append () for parameterless functions in completion

Dima-369:
See https://github.com/idursun/jjui/discussions/528#discussion-9474183


nickchomey:
Is it possible to append `(` for functions that do have parameters? eg typing `tag` and then tab would show `tags(`?

PR #532, opened 2026-02-14
fix(ui/git): update key for `jj git fetch --tracked`

PrayagS:
Noticed that both the default fetch and the tracked options are set to use the key `f`.

I'm assuming that's not intended. This PR updates it to use the key `t` similar to how `jj git push --tracked` is configured.


idursun (approved):
Thanks!

Issue #435, opened 2025-12-18
IP: Ability to load lua configuration.


vic:
I'm willing to put some effort on implementing this if we agree that this is something desirable. I'll try to explain what I mean and why I find it would be useful for jjui users.

First of all, I think the new lua bindings and lua-based custom-commands open new possibilities for customizing jjui behavior, far beyond what our old toml based custom-commands and leader-key enabled. So I'm quite happy that we have a scripting engine in jjui.

Sometimes I find myself working on several projects at once, and each project has its own conventions, at some I'd like jjui to behave differently, eg allow certain custom commands only on some projects because those custom-commands are useful only on them. Or, have different theme loaded so I can visually differentiate between jj workspaces (eg, when merging into main branch of a repo, using the fire theme allows me to visually know I should be pushing with care -- I use the same color-coding convention for example when openning production-DB connections where colors remind me to always work with transactions and more care). 

Currently I just copy my whole user-level configuration somewhere and set JJUI_CONF_DIR for each project. It works, and that's why I haven't raised an issue about this. 

`jj` itself has ways to configure depending on the directory you are at, via [scope tables](https://github.com/jj-vcs/jj/blob/main/docs/config.md#using---scope-tables) and I was thinking in the past of having something similar in jjui, so I could have per-project specific configurations.

But now that we are having Lua, I think we could use lua itself to import files (configuration fragments) conditionally. For me, it is much better option than having "imports"/"paritioning" like scoped tables. 

However I also value the fact that TOML is not a program, so I'm not proposing we to get rid of it. What I'm proposing is, if jjui finds a `config.lua` besides `config.toml`, the lua file could be loaded *after* the normal toml configuration has been loaded. 

The lua code can enhance and conditionally add configurations based on the working directory or any other condition the user feels right. 

People could also share small lua repos with their jjui custom-commands or extensions.

Anyways, if any of this makes sense the gist of it all is: Allow config.lua to be loaded and enhance jjui runtime config.

baggiiiie:
> What I'm proposing is, if jjui finds a config.lua besides config.toml, th

this is what popped in my mind when lua script was added, it's just like how i would config neovim. and currently, lua scripts are supported via plain text in toml file, which adds a little bit of friction to the configuring process

> each project has its own conventions, at some I'd like jjui to behave differently

if `config.lua` is used, i think this should be done within the lua script itself (e.g., "return empty table when current dir is xyz"), which is way cleaner than `scope table` imo.

[...]

PR #533, opened 2026-02-14
Unified Actions + Bindings system

idursun:
> [!NOTE]
> Diff is huge but mostly because of generated code for built-in actions and large chunks of removed leader/custom command/key handling code.

This PR replaces the legacy keybinding system (hardcoded `KeyMsg` handling, `[keys]` config, `[custom_commands]`, `[leader]` sequences) with a unified **actions + bindings** architecture. All keyboard input now flows through a single dispatch pipeline:

**KeyMsg -> Dispatcher -> Binding -> Action -> Intent -> Model.Update**

## Motivation

Previously, keybindings were hardcoded in each model's `Update` method via `key.Matches(msg, ...)` blocks. Custom commands, leader keys, and sequence overlays were separate systems with their own config formats and dispatch logic. This made it difficult to add new bindings, customize behavior, or expose actions to Lua scripts.

The new system centralises all input handling into a single dispatcher. Models no longer deal with key presses directly; they only respond to intents. This makes models simpler, input handling uniform, and the full set of actions discoverable and configurable.

### New Dispatch Pipeline (`internal/ui/dispatch/`)

- **`Dispatcher`**: Resolves key presses against scoped bindings, supporting both single-key and multi-key sequence bindings.
- **`Resolver`**: Extends the dispatcher to resolve all the way from key press to intent, checking operation overrides, the generated catalog, and configured Lua actions in order.

### Code Generation (`cmd/genactions/`)

Intent types in `internal/ui/intents/` are annotated with `//jjui:bind` comments that declare which scope and action they belong to, along with optional field mappings. A code generator scans these annotations and produces:

- **`catalog_gen.go`** (`internal/ui/actions/`): Maps action IDs to intent constructors with argument parsing.
- **`builtins_gen.go`** (`internal/ui/actionmeta/`): Action metadata (arg schemas, required args, owners) used for config validation help.

This eliminates manual boilerplate when adding new actions. Annotate the intent type and regenerate. It also makes built-in actions callable from Lua scripts through the shared action catalog.

To regenerate after changing intent annotations:

```bash
go run ./cmd/genactions
```

A staleness test ensures generated code stays in sync with annotations.

> [!NOTE]
> A `Makefile` with `build`, `run`, `test`, and `generate` targets would improve this workflow.

### Declarative Binding Configuration

All default key bindings are declared in `internal/config/default/bindings.toml`, separate from code. Each entry specifies `key` (or `seq` for multi-key sequences), `action`, `scope`, and optional `args`:

```toml
# Single key binding
[[bindings]]
key = "a"
action = "revisions.abandon" # built-in action
scope = "revisions"          # where this binding is active

# Multiple keys for the same action
[[bindings]]
key = ["up", "k"]
action = "revisions.move_up"
scope = "revisions"

# Multi-key sequence (replaces leader keys)
[[bindings]]
seq = ["g", "r"]
action = "ui.open_revset"
scope = "revisions"

# Binding with arguments
[[bindings]]
key = "r"
action = "revisions.rebase.set_source"
scope = "revisions.rebase"
args = { source = "revision" }
```

Scopes form a chain from innermost to outermost. The dispatcher walks the chain to find the first matching binding, giving inner scopes priority.

**User customization**: users add `[[bindings]]` entries in config to override or extend defaults. A merge strategy removes shadowed bindings by scope+key/seq, so users only specify what they want to change:

```toml
# Rebind abandon to 'x' instead of 'a'
[[bindings]]
key = "x"
action = "revisions.abandon"
scope = "revisions"
```

### Custom Actions in TOML and Lua

Users can define custom actions in TOML and bind them via `[[bindings]]`:

```toml
[[actions]]
name = "copy_diff"
lua = '''
local diff = jj("diff", "-r", context.change_id(), "--git")
copy_to_clipboard(diff)
'''

[[bindings]]
action = "copy_diff"
scope = "revisions"
seq = ["w", "y"]
desc = "copy diff"
```

This branch also adds first-class `config.lua` setup support:

- `config.lua` lives next to `config.toml`.
- `setup(config)` can register actions and bindings programmatically.
- New API names:
  - `config.action(name, fn, opts?)`
  - `config.bind({...})`
- `config.action(..., opts)` supports inline binding via `key`/`seq` + `scope`.
- `config.action(..., opts.desc)` and `config.bind({ desc = ... })` set per-binding help text.
- `config.action(..., opts)` enforces `key` and `seq` as mutually exclusive, and requires `scope` when either is provided.
- `config.bind({...})` currently supports `action`, `scope`, `key`/`seq`, and `desc` (binding `args` are not currently supported from `config.lua`).
- Lua `require(...)` resolves modules from the jjui config directory (`?.lua` and `?/init.lua`).
- `config.lua` and TOML actions share the same Lua VM, so Lua helpers defined in `config.lua` can be used by TOML actions.
- Global functions declared in `config.lua` can be called directly from `[[actions]].lua` scripts in `config.toml`.

Example:

```lua
function say_name(prefix)
  flash(prefix)
end

function setup(config)
  config.action("say hello to me", function()
    say_name("hello mister")
  end, {
    desc = "hello to me",
    seq = { "w", "p" },
    scope = "revisions",
  })

  config.bind({
    action = "ui.open_revset",
    key = "R",
    scope = "revisions",
  })
end
```

The status bar help now derives keys directly from active bindings in the current scope, so custom and rebound keys are reflected automatically without maintaining separate hardcoded lists.

A `bindings_profile` config option allows loading an alternative base bindings file.

### Configuration UX and Migration

- `jjui --config` opens (and creates if needed) `config.toml` in `$EDITOR`/`$VISUAL`.
- `jjui --config --migrate` migrates legacy `[custom_commands]` into `[[actions]]` + `[[bindings]]`.
  - Creates `config.old.toml` backup on first run.
  - Removes `[custom_commands]` from active config.
  - Appends generated `[[actions]]` / `[[bindings]]`.
  - Prints migrated entries and warnings.

## How can I test this branch?

1. Build and run:

```bash
go build ./cmd/jjui
./jjui
```

2. Open config and add a TOML action + binding:

```bash
jjui --config
```

```toml
[[actions]]
name = "say-hi"
lua = '''
flash("hello from config.toml")
'''

[[bindings]]
action = "say-hi"
key = "H"
scope = "revisions"
desc = "say hi"
```

3. Add `config.lua` next to `config.toml` and define Lua setup actions:

```lua
function setup(config)
  config.action("say hello to me", function()
    flash("hello from config.lua")
  end, {
    desc = "hello to me",
    seq = { "w", "p" },
    scope = "revisions",
  })
end
```

4. Verify plugin-style extension loading via `require(...)`:

```lua
-- ~/.config/jjui/plugins/my_extension.lua
local M = {}

function M.setup(config)
  config.action("say hello", function()
    say_name("hello world")
  end, { desc = "hello", seq = { "w", "h" }, scope = "revisions" })
end

return M
```

```lua
-- ~/.config/jjui/config.lua
local my_extension = require("plugins.my_extension")

function setup(config)
  my_extension.setup(config)
end
```

Technically, loading another Lua file and calling its `setup` from `config.lua` is enough to support plugins.

You can also define a global helper in `config.lua` and call it from TOML actions:

```lua
function helper(msg)
  flash(msg)
end
```

```toml
[[actions]]
name = "helper-demo"
lua = 'helper("called from TOML")'
```

5. Re-run `jjui` and verify both bindings (`H` and `w`,`p`) in the revisions view.
6. Use `internal/config/default/bindings.toml` as the source of truth for available scopes/actions while rebinding.
7. If you have legacy `[custom_commands]`, run migration and verify generated blocks:

```bash
jjui --config --migrate
```

## What Was Removed

The unified system replaces several separate mechanisms:

- **Leader keys** (`[leader]` config, `internal/ui/leader/`): replaced by sequence bindings (`seq` in `[[bindings]]`).
- **Custom commands** (`[custom_commands]` config, `internal/ui/custom_commands/`, `internal/ui/context/custom_*.go`): replaced by `[[actions]]` + `[[bindings]]`.
- **Hardcoded key dispatch** (`internal/config/keys.go`, `key.Matches` blocks in `ui.go` and operation models): models no longer handle `tea.KeyMsg` directly. All key-to-intent translation happens in the dispatch layer.
- **`[keys]` config section**: rejected at load time with a migration message to `[[bindings]]`.

#515

### DIFF
--- a/internal/config/migrate.go
+++ b/internal/config/migrate.go
@@ -29,7 +29,7 @@ func migrateWith(stdout, stderr io.Writer) int {
 	sourceFile := configFile
 	if _, err := os.Stat(backupFile); err == nil {
 		sourceFile = backupFile
-		fmt.Fprintln(stdout, "Old file found. Migrating from old file.")
+		fmt.Fprintf(stdout, "Migrating from '%s' to new format.\n", backupFile)
 	}
 
 	data, err := os.ReadFile(sourceFile)
--- a/README.md
+++ b/README.md
@@ -6,6 +6,11 @@
 
 If you are new `jjui`, have a look at [previously on jjui](https://github.com/idursun/jjui/discussions/443).
 
+> [!NOTE]
+> Trunk version is at v0.10 which is not released and contains breaking changes. See #533 for details.
+> If you want to migrate your configuration to the new version, try running `jjui --config --migrate` (best effort)
+> This should copy your existing configuration as `config.old.toml` and migrate it to a new `config.toml` file.
+
 ## Features
 
 ### Change revset with auto-complete
--- a/internal/config/bindings_schema_test.go
+++ b/internal/config/bindings_schema_test.go
@@ -121,18 +121,6 @@ lua = "print('x')"
 	assert.Equal(t, "print('x')", cfg.Actions[0].Lua)
 }
 
-func TestLoad_RejectsLegacyKeysSection(t *testing.T) {
-	content := `
-[keys]
-up = ["up", "k"]
-`
-
-	cfg := &Config{}
-	err := cfg.Load(content)
-	require.Error(t, err)
-	assert.Contains(t, err.Error(), "[keys] is no longer supported")
-}
-
 func TestLoad_CanonicalBuiltInEnumArgsValidation(t *testing.T) {
 	valid := `
 [[bindings]]
--- a/internal/config/loader.go
+++ b/internal/config/loader.go
@@ -95,9 +95,6 @@ func (c *Config) Load(data string) error {
 	if err != nil {
 		return err
 	}
-	if metadata.IsDefined("keys") {
-		return fmt.Errorf("[keys] is no longer supported; migrate to [[bindings]]")
-	}
 
 	// Decode only merge-managed array/table fields into a fresh struct so these
 	// collections are always read from file content, without carrying prior state.


nickchomey:
Very impressive! Can I suggest making a migration guide to help us switch to this?

idursun:
@nickchomey sure, it makes sense to provide a migration guide. But we can also do even better and provide a command like `jjui --config --migrate` and automate the whole thing. I'll give it a try.

baggiiiie:
wow always an incredible PR!

would u consider adding support similar to what i mentioned in https://github.com/idursun/jjui/issues/435#issuecomment-3883116076 ? or something similar to what @vic originally proposed?

vic:
I love this, @idursun. you are awesome!

idursun:
> wow always an incredible PR!
> 
> would u consider adding support similar to what i mentioned in [#435 (comment)](https://github.com/idursun/jjui/issues/435#issuecomment-3883116076) ? or something similar to what @vic originally proposed?

I am thinking about it but I still couldn't figure out the exact flow of it. 

Most of the folks will want to use TOML based syntax because it is easier and straight-forward.

It makes sense to load TOML first and then run the `config.lua` to allow config modifications. Power users will define their own Lua functions and use them in their Lua based action definitions.

However, I am almost 100% sure that some folks will reference their config.lua functions in the TOML actions too. This won't work because Lua scripts run as stand alone.

Maybe, we can say, by convention, `setup(config)` function in the `config.lua` is invoked at start up to change the configuration and everything else in the `config.lua` will be loaded while running actions at runtime so both TOML and `config.lua` based actions work. I need to spend some time to figure this out. It almost feels like a separate PR.

baggiiiie:
> by convention, setup(config) function in the config.lua is invoked at start up to change the configuration

yeah i think this offers a nicer experience. how lazyvim/neovim do it is quite intuitive (auto-load files under certain dir)

idursun:
I added `--migrate` and `config.lua` support and updated the PR description.

idursun:
Geez, I have just realised that we are one step away from having a plugin system with this `config.lua` change.

> yeah i think this offers a nicer experience. how lazyvim/neovim do it is quite intuitive (auto-load files under certain dir)

Technically we should be able to load another Lua file (`require("plugins.my_extension")`) and call its `setup` function from the `config.lua` and we are pretty much done in terms of supporting plugins.

```lua
-- ~/.config/jjui/plugins/my_extension.lua
local M = {}

function M.setup(config)
  config.action("say hello", function()
    say_name("hello world")
  end, { desc = "hello", seq = { "w", "h" }, scope = "revisions" })
end

return M
```

```lua
-- ~/.config/jjui/config.lua

local my_extension = require("plugins.my_extension")

function setup(config)
  my_extension.setup(config)
end
```

Hold my 🍺

baggiiiie:
quietly watching @idursun make things happen 👀

vic:
> I have just realised that we are one step away from having a plugin system with this `config.lua` change

yeah! exactly that was the reason behind me suggesting lua configs. :)

now just mix-in some existing lua-plugin loader and viola, community shareable jjui-configs.

idursun:
I think this PR is as good as it can get. After spending more than two weeks on it, I want to ship it one way or the other.

I am marking this ready for review and plan to fix stuff I missed in subsequent PRs. Merging this PR will mark the end of the 0.9 series.

CONTRIBUTING.md
```diff
@@ -58,11 +58,21 @@ This process helps ensure that:
...
+
+### Code Generation
+
+If you add or modify intent types with `//jjui:bind` annotations in `internal/ui/intents/`, you must regenerate the action catalog:
+
+```shell
+go run ./cmd/genactions
+```
+
+This produces `catalog_gen.go` and `builtins_gen.go`. A staleness test will fail if generated code is out of sync with annotations.
```
vic, line 71: +1 on having this run manually for now, so good we have developer comments about this and out-of-sync checks.

nix/partitions.nix
```diff
@@ -17,9 +15,7 @@
       { inputs, ... }:
       {
         imports = [
-          inputs.treefmt-nix.flakeModule
```
vic, line 20: +1 on removing treefmt.

I see the nix CI step still verifies that the jjui app is built correctly and that is enough for me.

internal/config/migrate.go
```diff
@@ -0,0 +1,341 @@
...
+	"os"
+	"path/filepath"
+	"regexp"
+	"sort"
+	"strings"
+
+	"github.com/BurntSushi/toml"
+)
+
+// Migrate converts legacy [custom_commands] config to [[actions]] + [[bindings]].
```
vic: awesome! kudos for this migration tool

cmd/jjui/main.go
```diff
@@ -143,23 +152,43 @@ func run() int {
...
 		return 1
 	}
 
+	if err := scripting.InitVM(appContext); err != nil {
+		fmt.Fprintf(os.Stderr, "Error initializing Lua VM: %v\n", err)
+		return 1
+	}
+	defer scripting.CloseVM(appContext)
+
+	if luaSource, err := config.LoadLuaConfigFile(); err != nil {
```
vic, line 169: Yaay! Thanks for lua configs.

Adda0 (approved):
The Nix changes LGTM. If you found the formatter more of a bother than useful, then by all means, remove it.

Out of curiosity, what did you find problematic about treefmt? Any particular reason, or just generally too strict and bothersome to maintain the other files, other than just Go code, like the markdown, formatted?

Great changes, otherwise.

idursun:
@Adda0 
> Out of curiosity, what did you find problematic about treefmt?

I pushed bindings.toml file formatted in an easily scannable way but nixfmt failed the build, which then I disabled it for TOML files but the build failed again. At that point, I didn't know how to fix it and removed it altogether. I didn't like the fact that the build failing even though everything passing on my machine.

idursun:
I fixed a couple of issues around Lua scripts and a few issues in the migration command. 

I couldn't find anything else **but I am sure there will be one as soon as merge this!**

baggiiiie:
nice work!

just pulled down main and tried out, migration didn't work perfectly for me, still have quite a bit of manually adjustment, i'll try to find out why and post some fixes if i can

it's a bummer that `scope` only allows string, would be great if it supports array. what do you think @idursun ?

---

now pressing `p` doesn't work in preview, i had a quick look at the codebase, seems like it's due to `p` only binds to `revisions` scope. similarly, `@` isn't available in rebase/squash

for this specific case, we could add another line in `bindings.toml` to allow `p` in `revisions.details`. 

in the long run, i think it would be great to support array for scopes, for better `plugins/custom_command.lua` UX
or we could have inheritance, e.g., `revisions.details` automatically inherit bindings from `revisions`, unless being overwritten

idursun:
> for this specific case, we could add another line in bindings.toml to allow p in revisions.details.

Yes, for the time being I prefer duplication over code complexity. In reality, most users will spend some time with configuration and probably hate the duplication during that time but I will hate the code complexity almost every day.

baggiiiie:
> In reality, most users will spend some time with configuration and probably hate the duplication during that time but I will hate the code complexity almost every day.

well said!

PR #541, opened 2026-02-20
fix: add back `esc` to clear checked revisions

baggiiiie:
Previously, the `esc` key would clear the checked revisions. Adding back
this functionality here


idursun:
@baggiiiie Do you find the current codebase easier or harder to navigate?

baggiiiie:
@idursun as of right now im still getting used to the codebase. at this specific moment it's not a bit against my muscle memory, but i think this just takes some getting used to. it's more well-structured for sure tho!

the way it generates code is very cool! feels like convex, or sqlc. really love it! the changes you've brought in (like this, and the lua support), make this project more and more interesting to work on as a contributor!

PR #546, opened 2026-02-23
move docs to astrojs/starlight site

### LINKED ISSUE #543: Move website to astrojs for better sidebar and content organization
> Btw, @idursun, this weekend I will work on giving us a better webpage, I see the current sidebar is quite big now. I intend to use astrojs, just like I have for the documentation of my den lib: https://den.oeiuwq.com. The advantage is that we will have better navigation and ways to have pushed migrating guides, etc.  

 _Originally posted by @vic in [#536](https://github.com/idursun/jjui/issues/536#issuecomment-3937645277)_

vic:
regen pnpm lock

closes #543


vic:
Site works locally but I'm investigating why deployment has no css assets.

vic:
I was missing this: https://github.com/idursun/jjui/commit/51600d604583c2541f9630eb2b5cf52c05cf431e

because the site is not at top-level.

vic:
I will also fix all links now.

idursun:
Thanks @vic! This looks awesome!

So with the new setup, can we publish only the documentation for v0.9.12 but also start updating/writing the documentation for v0.10 without publishing it until it is released?

vic:
@idursun, currently the site is published only from the docs branch. We can have WIP branches forked from docs branch but they wont be published, so you'd have to `pnpm run dev` to see the docs. And once merged the gh-pages action publishes the site. 

Another way would be to have more sections for upcoming documentation, etc. but that would be visible for people. Anything in the docs branch is published, if you want docs not to be published until a version is released, I'd suggest we keep other branches like docs/upcoming or docs/v0.11.0, and once they become *the* docs branch they'll be published. Having these docs/version branches would also help keeping docs for historical old-versions.

idursun:
> Another way would be to have more sections for upcoming documentation, etc. but that would be visible for people.

I don't mind publishing current documentation at all. 

What I actually want is to be able to work on the documentation of the current version while still having documentation of a released version published. Sections make a lot of sense here. 

My brain is not working today but I want the equivalent of this: 
image

idursun:
Maybe we don't need to complicate things and just do what you suggested, `docs` is published and `docs/v0.10` is the one we work on and it gets merged to `docs` and gets published once it is released. I am also OK with that.

vic:
I believe "versioned docs" might be possible with astro.build, given it is focused on documentation sites. I haven't read how to do this, tho. Maybe later we can have that drop-down you want.

vic:
For reference, first search result:

https://starlight-versions.vercel.app/guides/about-versioning/

demo:

https://starlight-versions-example.vercel.app/

PR #550, opened 2026-02-24
feat: add change_offset in jj log prefix template for divergent revisions

### LINKED ISSUE #545: Divergent revision parsing depends on jj log template
## Description
[I don't use ids in the log template](https://github.com/idursun/jjui/issues/76#issuecomment-2859249010) (jjui is so good I rarely need them).
Ran into a parsing error when trying to interact with divergent changes.
Worked around it by prepending divergent commits with the jj v0.37 change-id-with-offset (jjui could potentially start using it in it's log prefix?).
Just wanted to share the workaround in case anyone else runs into the same error:
```text
[template-aliases]
format_short_commit_header(commit) = '''
  separate(" ",
    if(commit.divergent(), format_short_change_id_with_change_offset(commit), ""),
    commit_timestamp(commit).ago(),
    format_short_signature(commit.author()),
    commit.bookmarks(),
    commit.tags(),
    commit.working_copies(),
    format_commit_labels(commit),
  )
'''
```

Use a jj template alias like this:
```text
[template-aliases]
format_short_commit_header(commit) = '''
  separate(" ",
    commit_timestamp(commit).ago(),
    format_short_signature(commit.author()),
    commit.bookmarks(),
    commit.tags(),
    commit.working_copies(),
    format_commit_labels(commit),
  )
'''
```

Try to expand or abandon a divergent change.

## Expected Behavior
The divergent change is expanded or abandoned.

## Actual Behavior
```
Error: Failed to parse revset: Syntax error
Caused by:  --> 1:4
  |
1 | 20 hours ago
  |    ^---
  |
  = expected <EOI>, `|`, `&`, or `~`
```

## Specifications

- Platform: NixOS
- JJUI Version: main-720520c
- JJ Version: 0.38.0

baggiiiie:
Previously, when abandoning a divergent revision, a full ChangeID is used by detecting double question marks at the end of a change ID. This has been deprecated since the jj 0.36.0 release, and since then we have introduced `_PREFIX` template in `jj log` to fetch divergent info.

Replace fragile segment-scanning logic that reconstructed full change IDs (ending in ??) with a simpler approach: the jj template now outputs change_id/offset direclty in the prefix for divergent revisions. And when abandoning a divergent revision, we simply `jj abandon $ChangeID/offset`

This eliminates `IsConflicting()`, removes the need to fall back to CommitId, and simplifies both the parser and `GetChangeId()`.

Fixes #545

### DIFF
--- a/internal/jj/commands.go
+++ b/internal/jj/commands.go
@@ -53,7 +53,7 @@ func Log(revset string, limit int, jjTemplate string) CommandArgs {
 		template = jjTemplate
 	}
 	prefix := fmt.Sprintf(
-		"stringify('%s' ++ separate('%s', change_id.shortest(), commit_id.shortest(), divergent))",
+		"stringify('%s' ++ separate('%s', change_id.shortest() ++ if(divergent, \"/\" ++ change_offset), commit_id.shortest(), divergent))",
 		JJUIPrefix, JJUIPrefix)
 	template = fmt.Sprintf("%s ++ ' ' ++ %s", prefix, template)
 	args = append(args, "-T", template)
--- a/internal/jj/commit.go
+++ b/internal/jj/commit.go
@@ -1,7 +1,5 @@
 package jj
 
-import "strings"
-
 const (
 	RootChangeId = "zzzzzzzz"
 )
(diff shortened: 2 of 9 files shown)


baggiiiie:
@idursun i have thought about a few different approaches, hope below learnings of mine help you with your experiment:
- two threads, decouple visual rendering and data fetching; but it takes extra effort to sync two threads, and might cause issue with oplog
- parsing data with visual information should be avoided, since users have unlimited combination and creativity with jj/jjui config
- current templating approaching with `jj log` is the best i could think of. could probably be improved with null byte delimiter, and include bookmark info if we hit edge cases with user configs 
- or putting metadata in a separate row, and hide it in rendering. might be a simpler approach when we have too much metadata

PR #551, opened 2026-02-24
fix: make bookmark input visible when user has custom template-aliases

### LINKED ISSUE #519: Typed text is not visible when adding bookmark
## Description

When pressing `B` to add a new bookmark I don't see the string I am typing. First I thought that this is a theme issue and I might have the same `bg` and `fg` color set for this particular feature (I have not). 
But in the video in the README it can be seen that cursor shows up while typing the new bookmark name and the git sha moves further to the right. This does not happen  for me. I can type though and use backspace for correction, etc. When I press Enter then the bookmark is created successfully.

## Expected Behavior
Typed text is shown.

## Actual Behavior
Typed text is not visible

## Specifications

- Platform: macOS  26.2
- Version: jjui 0.9.10

baggiiiie:
The bookmark text input is invisible when users had custom jj `template-aliases` that changed how commit_id is rendered. The renderer matched `segment.Text == CommitId` exactly, which failed when the visible commit_id format differed from the prefix-extracted `commit_id.shortest()`.

Changes:
- Use strings.HasPrefix instead of exact equality to match the commit_id segment, handling cases like id.short(12) vs id.shortest()
- Add fallback: if no commit_id segment is found at all (e.g. user template omits it entirely), append the bookmark input at end of line instead of silently dropping it
- Track insertion with a flag to prevent double-rendering

Fixes #519, also works for the case where the user has no ChangeID/CommitID, see #545

To reproduce, apply user's `jj template-aliases` in #519 and #545, and press `B` to start setting bookmark.

### DIFF
--- a/internal/ui/revisions/displaycontext_renderer.go
+++ b/internal/ui/revisions/displaycontext_renderer.go
@@ -502,13 +502,19 @@ func (ir *itemRenderer) renderLine(tb *render.TextBuilder, line *parser.GraphRow
 
 	// Render segments
 	beforeCommitID := ""
-	if ir.op != nil {
+	if ir.op != nil && line.Flags&parser.Revision == parser.Revision {
 		beforeCommitID = ir.op.Render(ir.row.Commit, operations.RenderBeforeCommitId)
 	}
 
+	// A flag to track whether beforeCommitID was successfully inserted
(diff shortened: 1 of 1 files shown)


baggiiiie:
@idursun thanks for the trust!

PR #556, opened 2026-02-26
feat(flash): command history

### LINKED ISSUE #243: IP: Have a way to see previously closed flash notifications.
This is subject to it making sense to maintainers and it being something of value for jjui users. Please vote 👍 / 👎 .

from [#240 comment](https://github.com/idursun/jjui/issues/240#issuecomment-3133123518) about making the output of commands sticky, and [later](https://github.com/idursun/jjui/issues/240#issuecomment-3133541181) about instead of making non-errored output sticky, have a way to see output of already closed flash messages.

**What are you planning to implement?**

Show a window with the list (latest first) commands run with their respective output.

**Proposed Approach**

- How will this feature work? For example, describe how the user would interact with jjui using this new feature.

A new shortcut will open our center stacked view with a list of previously closed flash messages.

Each "row" has a success or error emoji, followed by exit status (int), the timestamp and command that was run colored either green or red according to error status. Followed by the lines of the command output.

You can move as usual up or down using j/k  up/down and close using esc. No other key is handled during this view.   

shortcut proposal: `F` (flash history).  (current lowercase `f` is ace jump) 

- How will this integrate with or impact existing features?

We will be able to show the output of all commands run by jjui on behalf of the user, except for those of `sh exec` or `jj exec` since these are given complete IO control and jjui does not intercepts their output in any way.

**Implementation Details**

Output of commands is not stored in disk, and is lost when jjui terminates. 

**Testing Strategy**
How do you plan to test this implementation?
- Manual testing approach

**Alternatives Considered**
What other implementation approaches have you considered? Why did you choose this particular approach?

An alternative is not to have another window for flash messages but be able to see the output at `oplog` (not sure about this) 

**Breaking Changes**
N/A

**Questions for Discussion**
Do you would like for jjui to have this feature?

idursun:
I was trying to simplify `status.go` so decided to move the running command information to the flash system so that the command and its output are presented close to each other. (which I think is much better overall).

Then I noticed I am not far off from implementing command history. Here's how it looks at the moment:

https://github.com/user-attachments/assets/76b024f9-e45d-49d7-bbdf-d8a72513ed7d

I bound opening command history to `W` (which is up for debate). I think it looks visually OK but I might be biased.

I've been using this for the last couple of days and I already like the experience. I am opening the PR to collect some feedback. 

fixes #243

### DIFF
--- a/internal/ui/commandhistory/model.go
+++ b/internal/ui/commandhistory/model.go
@@ -101,6 +101,9 @@ func (m *Model) ViewRect(dl *render.DisplayContext, box layout.Box) {
 	y := area.Max.Y - 1
 	maxWidth := area.Dx() - 4
 
+	rest, _ := box.CutBottom(1)
+	dl.AddDim(rest.R, render.ZOverlay)
+
 	for _, item := range m.window() {
 		content := m.renderEntry(item.entry, maxWidth, item.selected)
 		w, h := lipgloss.Size(content)


seh:
> I am opening the PR to collect some feedback.

Having watched the video a couple of times, I found the overlays to be distracting and confusing, with them stacking up and going away, making it difficult to tell what it all "meant". Perhaps if I were controlling it with the keyboard myself I would be able to interpret the effects of navigating and deleting the entries.

I expected to see a keybinding for _repeating_ an operation, though I suppose that there aren't too many operations performed in _jjui_ that warrant repeating exactly later.

idursun:
@seh @baggiiiie Thanks for the feedback!

> I expected to see a keybinding for repeating an operation, though I suppose that there aren't too many operations performing in jjui that warrant repeating exactly later

Exactly. Never occurred to me that this could be useful, I don't think it would ever. I'd recommend repeating the command using the operation again instead of running it from the command history.

Command history is useful when you want to see the output for a little bit longer (e.g. which branches are deleted/created, PR link of a git push, etc)

> i think a different color to indicate current selection migth be better than double border.

I will have a look at this. There's too much green at the moment, toning it down might look better.

idursun:
Ok, I had tried dim effect previously (before the bubbletea v2 migration) and it didn't look great back then. Specifically, `faint=true` style didn't apply uniformly to all screen cells but seems like it this is fixed in v2. It looks like this now and I think this is even better (notice that the rest of the UI is washed out)

image

I will push new commits to this branch but this look only appears once the v2 branch is merged.

locksten:
I really like the running command indicator being more prominent – it was easy to miss before.
The dimmed background definitely looks nice.
I thought keeping the titles right-aligned while scrolling through history would be nicer, though after making a mock screenshot I'm not sure 🤔. It might be better to keep it consistently left-aligned.

jjui align

nickchomey:
This looks really nice. Like others said, at first, I couldnt figure out the use case - you can't replay them. But why would you even want to? Its nice being able to see the history, particularly for when you want to inspect the command that was run for a complicated action.

idursun:
Thanks @nickchomey @locksten 

I think I am going to ship this as is. 

One nice addition to this might be copying the output to the clipboard. (maybe pressing `y`) but you can still do this today by shift dragging. We can think about it later.

nickchomey:
Yeah, I'd say ship it. 

If you do add a copy feature, I think it should only copy the command.

PR #558, opened 2026-02-27
Update to bubbletea v2

idursun:
This PR upgrades Bubble Tea to v2.

Most of the migration work had already been done during the RC2 update, and we’ve gradually removed v1-specific usage over time. As a result, the final migration was straightforward, even though it touched many files. Behaviour appears unchanged, and this PR does not yet adopt new v2 features; we can evaluate those in follow-up work.

I also removed the experimental `cellbuf` package usage in favour of `ultraviolet`. `Rect`/`Rectangle` types have been moved into the `layout` package, and we now align with the existing ultraviolet approach for similar functionality. This helps keep internals encapsulated and avoids exposing implementation details across the app.

Stacked on top of #556


idursun:
I used this branch for a day now and seems to be working just fine. The best way to find out is to ship it as there are quite a few users using the `main` branch daily.

PR #564, opened 2026-02-28
feat: add `(` to functions when autocompleted

nickchomey:
I finally decided to use AI to help me make changes to jjui (along with using jjui itself to navigate the different branches etc to test out new functionality)!

A small first PR: when you autocomplete `ta` it'll fill `tags(` with the opening `(`. functions like `all` still autocomplete as `all()` 

 It was, of course, an easy and seamless experience. Looking forward to contributing more in the future!

### DIFF
--- a/internal/jj/source/source.go
+++ b/internal/jj/source/source.go
@@ -26,10 +26,13 @@ type Item struct {
 	HasParameters bool
 }
 
-// DisplayName returns the item name, appending "()" for parameterless functions.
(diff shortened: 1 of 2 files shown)


nickchomey:
One question: should it show the opening parenthesis in the list of functions? Alternatively, should the `()` be removed from `all()` etc...?

image

nickchomey:
ah, yes. of course. Fixed in #567

PR #566, opened 2026-03-01
DO NOT MERGE: Add line wrapping by adding a PTY

skorokithakis:
This is just an idea for how line wrapping could work. The PR creates a PTY for the preview pane and disables the jj pager, sets TERM=dumb to suppress queries, and sets SoftWrap to true so it wraps.

I'm just submitting it for inspiration, though it does work. I don't have an intention of it being merged.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -14,6 +14,11 @@ require (
 )
 
 require (
+	github.com/creack/pty/v2 v2.0.1
+	golang.org/x/term v0.38.0
+)
+
+require (
 	github.com/charmbracelet/colorprofile v0.4.2 // indirect
 	github.com/charmbracelet/ultraviolet v0.0.0-20260205113103-524a6607adb8
 	github.com/charmbracelet/x/termios v0.1.1 // indirect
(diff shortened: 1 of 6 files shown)


baggiiiie (commented):
i see what you're trying to do is wrap long lines in the preview pane, by using pty and making jj believe it is writing to a terminal of the pane's size.

to deal with line wrapping in preview, i think a better approach *might* be adding a preview setting like `preview.wrap = bool` to toggle `viewport.Model{SoftWrap: ...}`.

`$preview_width` could be used where jj can format to width, instead of forcing TTY emulation. tools like delta could leverage it. not sure about jj's built-in diff view tho

skorokithakis:
I don't think delta accepts any setting to force word wrapping, and it doesn't respect $COLUMNS either :/ That's probably something to be fixed on delta's side, though.

baggiiiie:
yeah delta's tricky. im looking for a better diff tool, and not enjoying difft either

tho regarding preview width, do u think this is, or closed to, what u want?

https://github.com/user-attachments/assets/aa29a0d1-ca0c-496b-9ab1-611caa8ca2a2

skorokithakis:
Yes, this is much better. I prefer inline rather than side-by-side, but the only reason I'm still using lazygit is that it shows a better diff, and I'd like to move completely away.

idursun:
Well, I am using the following if I want a diff with file tree. However, `diffnav` is not doing soft wrap either.

```lua
  config.action("show diff in diffnav", function()
    if not change_id or change_id == "" then
      flash({ text = "No revision selected", error = true })
      return
    end

    exec_shell(string.format("jj diff -r %q --git --color always | diffnav", change_id))
  end, { desc = "show diff in diffnav", key = "ctrl+d", scope = "revisions" })
```

I don't think faking a PTY is the way to solve it though. I think soft wrapping is a presentation problem and should be solved on the `jjui` side alone.

I did some work a while back around the time when I refactored the preview window to use the new render package instead of viewport and I had also removed it from the diff view in a follow up change. I will try reviving that work and see if soft wrapping can be added. 

We have just migrated to bubbletea v2 and it is using `ultraviolet` under the hood, and `uv.StyledString` has built-in wrapping support which can actually come in handy for this work.

skorokithakis:
That sounds good, thanks! I will close this now, as your jjui-side idea sounds good.

PR #567, opened 2026-03-01
fix: remove parentheses from functions list, but add when autocomplete

nickchomey:
image


baggiiiie (commented):
this is neat!

idursun (approved):
Thanks!

PR #568, opened 2026-03-01
fix(ui): make expanded status help scrollable with navigation scope

baggiiiie:
Currently, in `revisions` view the expanded status menu is too long, and
when the screen size is too small and expanded menu displays as one
column, it covers the whole viewport and is not scrollable.

Add a new ui.status_expanded scope with bindings for line, half-page,
and page navigation (ctrl+p/n, ctrl+u/d, pgup/pgdown), and map them to
new StatusScroll intents.

Track expanded-help viewport state (offset, visible, total) in status
rendering, clamp scrolling safely, and render a directional indicator
(up/down/both) in the mode label to show whether more content exists
above or below.

Update dispatch scope ordering so ui.status_expanded takes precedence
when the status overlay is open, while keeping status help output
focused on primary/always-on scopes.

Also add regression coverage for scope precedence so overlapping keys
resolve to status scrolling before preview scrolling when both scopes
are active.

| before | after | 
| -- | -- |
| image | image |


baggiiiie:
i'm thinking about adding a search function to status.go, UX wise it'd be the same as other search like in Git and Bookmarks, to address https://github.com/idursun/jjui/pull/503#issuecomment-3916195523 by @academician

idursun:
Thanks, this looks reasonable, but I'd like to step back and reconsider the direction before adding more complexity.

The expanded status view was originally designed to show all key bindings, and previously we were intentionally curating which ones appeared there. With the new system that limit is gone, which is what's causing the overflow problem. But this feels like an edge case — most of the time the screen space is sufficient — and I'm not sure we should optimise for it by introducing new scrolling mechanisms.

For the narrow screen problem specifically, I think the right answer is simply to limit how many items are shown and let the user widen their terminal to see the rest. That's a natural and expected behaviour. Introducing a scroll mechanism for this case feels like too much.

More broadly, I think we need to rethink the status view's role. It's already doing too many things: quick search, find file, shell execution, jj command runner. I don't want to add scrollable key binding navigation with search on top of that. Instead, I'd prefer to move each of these concerns into their own dedicated model implementations.

The same thinking applies to the preview panel — I don't love that it is currently controlled by bindings defined in other namespaces (e.g. revisions, oplog, etc). A better approach would be to have a focus mechanism, where the user explicitly focuses into the preview or status panel and interacts with it there.

I'd like to step back and think a bit differently without committing to a specific solution.

baggiiiie:
@idursun good call, i agree with the role of status.go is getting bloated. this is not necessarily helping that. 

what kind of focus mechanism are u thinking about? e.g., pressing `tab` to sort of "go in" the preview mode and interact with it (scrolling etc)? 



for status.go im drawing some inspiration from [Helix](https://helix-editor.com/)

https://github.com/user-attachments/assets/919d0120-bf0a-4bd8-9eb3-401cd9ed159f

nickchomey:
> The same thinking applies to the preview panel — I don't love that it is currently controlled by bindings defined in other namespaces (e.g. revisions, oplog, etc). A better approach would be to have a focus mechanism, where the user explicitly focuses into the preview or status panel and interacts with it there.

We can currently use keyboard shortcuts to seamlessly move through the commit log/graph and details view, and scroll the preview mode without doing anything else. If we had to press something to activate preview scrolling, that would be a significant step back in UX.

baggiiiie:
> If we had to press something to activate preview scrolling, that would be a significant step back in UX.

from user level, stick to the name `preview`, it's a quick way to preview a revision, merely a visual layer of the revision. to me that's completely fine. i only use preview to have a quick visualization of something in the revision. for more comprehensive diff view, `d` is just one key stroke away. 

on the same note, it's due to this reason, im not bothered by the offset of preview pane being refreshed triggered by auto refresh, when im still looking at the preview. in fact, i think it prompts me to do things properly: if im staring at a diff for so long that it auto refreshes, it's probably more optimized to use a diff view

curious about how other users use it tho


> I don't love that it is currently controlled by bindings defined in other namespaces (e.g. revisions, oplog, etc). 

i think it comes down to how to perceive this preview mechanism. if it's treated as an extended ability (or a visualization layer) of revisions/oplog, having its key bindings defined in revisions/oplog isn't too bad. we could definitely abstract it out to an independent component handling its own key bindings, while still keeping its nature of being an extension of revisions/oplog

if we wanna extract preview to be an independent thing of its own, having its keybinding defined in other namesapces will probably cause more harm than good

baggiiiie:
hi @idursun @nickchomey , i opened an issue to discuss this and other related topics. closing this PR now

nickchomey:
> i only use preview to have a quick visualization of something in the revision. for more comprehensive diff view, d is just one key stroke away.

Amazing how different people's workflows can be! I ONLY use preview. Literally never use full diff. 

Though, I use both normal preview and details preview differently. I set up details preview to use 2 column diffs viw difft, but normal preview is just normal diff. This is because difft can be slow when there's lots of diffs (such as an entire revision/commit).

PR #572, opened 2026-03-02
Refactor/Lua bridge

idursun:
I was testing the new Lua scripting system and noticed a few consistency gaps, so I aligned the action model around generated intents. The naming had drifted over time: some closeable flows used `start_*`, some used `open_*`, and some had no prefix at all even though they opened something that could be closed. I standardised these to `open_*` for closeable operations/dialogs, because that naming naturally matches the lifecycle and makes scripted control flow clearer.

Previously, we also had handcrafted Lua bridge functions for a subset of operations, and those helpers implicitly waited so they could be used directly in `if` blocks. That handcrafted list will always drift now that Lua actions are auto-generated from intents, so it no longer reflects the real system surface. This change removes those custom bridge entrypoints and keeps equivalent behaviour through explicit wait helpers.

To preserve practical behaviour, two helpers are now the control-flow primitives: `wait_close()` and `wait_refresh()`. `wait_close()` lets scripts branch on close/apply outcomes, and `wait_refresh()` matches the fact that many internal commands refresh state after execution. I also aligned apply paths in closeable dialogs/operations (including git and bookmarks) to emit applied-close semantics, so script behaviour is consistent and predictable across the UI. Overall, this makes both the UI behaviour and the Lua script layer more controllable for multi-step workflows.

I also cleaned up binding order and removed duplicated bindings encountered during this refactor.

Here's an example for a potential use of the current Lua scripting. The following action will enter inline describe mode, let you type something, and when you press alt+enter, it will create a new commit and open inline describe on it. this will continue until user hits `esc` 

```lua
config.action("new + inline describe loop", function()
  while true do
    revisions.open_inline_describe()
    if not wait_close() then
      break
    end
    wait_refresh()
    revisions.new()
    wait_refresh()
    revisions.jump_to_working_copy()
  end
end, { desc = "new + inline describe loop", seq = { "w", "c" }, scope = "revisions" })
```

### DIFF
--- a/internal/ui/bookmarks/bookmarks.go
+++ b/internal/ui/bookmarks/bookmarks.go
@@ -370,7 +370,7 @@ func (m *Model) handleIntent(intent intents.Intent) tea.Cmd {
 		if !ok {
 			return nil
 		}
-		return m.context.RunCommand(selected.args, common.Refresh, common.Close)
+		return m.context.RunCommand(selected.args, common.Refresh, common.CloseApplied)
 	case intents.BookmarksFilter:
 		filter := string(msg.Kind)
 		if filter == "" {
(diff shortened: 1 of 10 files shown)


baggiiiie:
seems like there's a breaking change, `inline_describe` is changed to `open_inline_describe`

idursun:
Yes, everything that’s closeable starts with open now. 

0.10 is full of breaking changes but given that lua script support in 0.9 was experimental, this might not count one.

PR #571, opened 2026-03-02
fix(git): prevent panic when opening git menu with no selected revision

baggiiiie:
Root cause:
- `revisions.SelectedRevisions()` can call
  `jj.NewSelectedRevisions(m.SelectedRevision())` when no row is selected.
- `m.SelectedRevision()` can be nil.
- `SelectedRevisions` previously kept nil entries, so
  `git.NewModel/createMenuItems` dereferenced `commit.GetChangeId()` on
  a nil *jj.Commit and panicked.
- when the jj repo is big, initial load/refresh might not have produced
  rows before `revision.SelectedRevision()` is called.

Fixes:
- Filter nil *Commit values in `jj.NewSelectedRevisions` before storing
  them.
- Keep regression coverage for git model construction with a nil
  selected revision (no panic).
- Add a constructor unit test proving nil revisions are dropped.
- createMenuItems only iterates non-nil commits after construction
  filtering.
- this removes the invalid dereference path while preserving existing
  behavior for valid selections.

Closes #570

Issue #570, opened 2026-03-01
nil pointer derefence


SandaruKasa:
## Description

Maybe similar to #287?

## Steps to Reproduce the Problem

I run into this crash like once a month maybe.

1. cd big_repo
1. `jj ui` (I have it as an alias to `jjui`)

I think it only happens when I do this for the first time during a particular boot. But I may be wrong.

Doesn't really bother me, but I thought it'd be worth filing it anyways, so that at least it's catalogued. Amazing program, by the way.

## Expected Behavior

jjui launches

## Actual Behavior

```
Caught panic:

runtime error: invalid memory address or nil pointer dereference

Restoring terminal...

goroutine 1 [running]:
runtime/debug.Stack()
        runtime/debug/stack.go:26 +0x5e
runtime/debug.PrintStack()
        runtime/debug/stack.go:18 +0x13
github.com/charmbracelet/bubbletea.(*Program).recoverFromPanic(0xc0001fe3c0, {0x746c80, 0xafee50})
        github.com/charmbracelet/bubbletea@v1.3.10/tea.go:847 +0xac
github.com/charmbracelet/bubbletea.(*Program).Run.func2()
        github.com/charmbracelet/bubbletea@v1.3.10/tea.go:638 +0xe6
panic({0x746c80?, 0xafee50?})
        runtime/panic.go:783 +0x132
github.com/idursun/jjui/internal/ui/git.(*Model).createMenuItems(0xc000580008)
        github.com/idursun/jjui/internal/ui/git/git.go:666 +0xfd
github.com/idursun/jjui/internal/ui/git.NewModel(0xc000318000, {{0xc00008a278?, 0xc0001f09c0?, 0x1?}})
        github.com/idursun/jjui/internal/ui/git/git.go:428 +0xa8f
github.com/idursun/jjui/internal/ui.(*Model).handleIntent(0xc0004b0008, {0x886f80?, 0xb2fb80?})
        github.com/idursun/jjui/internal/ui/ui.go:625 +0x92b
github.com/idursun/jjui/internal/ui.(*Model).Update(0xc0004b0008, {0x780ee0, 0xc0003106f0})
        github.com/idursun/jjui/internal/ui/ui.go:237 +0x74f
github.com/idursun/jjui/internal/ui.(*wrapper).Update(0xc000466b60, {0x780ee0?, 0xc0003106f0?})
        github.com/idursun/jjui/internal/ui/ui.go:746 +0x5b
[...]

PR #573, opened 2026-03-02
fix(ui): guard against nil commit in OpenBookmarks intent

baggiiiie:
SelectedRevision() can return nil when the cursor is out of bounds (e.g., during initial load of a large repo, as mentioned in #570 ). This nil was passed directly to bookmarks.NewModel, causing a nil pointer dereference at bookmarks.go:563.

Add an early return when SelectedRevision() is nil, consistent with the existing nil-filtering fix in NewSelectedRevisions for the git module 

related to #570 and #571


idursun (approved):
Thanks!

PR #575, opened 2026-03-02
feat(diff): word wrap

### LINKED ISSUE #565: jjui does not wrap diffs to viewport
## Description

When viewing diffs, jjui's preview does not wrap long lines. I'm using delta as my diff viewer, but it seems that jjui does not set the viewport width.

1. Use delta as your diff viewer
1. View a file with long lines
1. The end of the lines gets cut off

## Expected Behavior

Lines should wrap.

## Actual Behavior

They do not.

## Specifications

- Platform: Ubuntu 24.04
- Version: 0.9.12

idursun:
I dug out the old soft wrapping code and adapted for the new `ultraviolet` package that came with bubbletea v2.

I’ve removed `viewport` from the diff view and replaced it with `uv.ScreenBuffer` and `uv.StyledString`, which gives us built-in wrapping. 

The diff now has two explicit modes: a normal no-wrap mode and a wrapped mode. 
* When entering wrapped mode, we precompute the soft-wrap layout data and then keep it in sync with width changes, so scrolling and bounds remain accurate. 
* Rendering in wrapped mode is done directly into the visible view area via `StyledString`. Wrap can be toggled with `w` in the diff view.

fixes #565


idursun:
I am going to merge this for now but feel free to comment to this PR if you see something that doesn't look right.

PR #576, opened 2026-03-03
Add pgup/pgdown keybindings to evolog and details

academician:
I've wanted to be able to navigate by pages in jjui in more views for a while; this PR does that for the details and evolog views. It is mostly inspired by (ie, copied from) the same behavior in the oplog view.

I don't love that I'm copy/pasting code here, of course. It feels like all this list navigation behavior is pretty similar between oplog, evolog, details, git, etc, and should be unified. But I wasn't sure where it should go.


idursun (approved):
Hey, thanks!

Duplication is fine as it shows the need for a better abstraction. You can always follow this PR up with your suggestion of a better abstraction.

PR #577, opened 2026-03-03
feat(lua): make default actions overridable and expose jjui.builtin.*

idursun:
Second episode in kicking the tyres of the Lua scripting system.

I wanted to change the default behaviour of `revisions.toggle_select` action to move the cursor down instead of jumping to the parent. Even though we allow users to define custom actions with the same built-in action names, we always resolve the built-in function first.

The following never gets called because built-in takes precedence.

```lua
config.action("revisions.toggle_select", function() 
  flash("toggle")
end)
``` 

I fixed that by changing the resolution order but then we lose access to the built-in actions and there's no way to toggle selection via Lua script.

To address that, I introduced `jjui.builtin.*`, which always points to built-in actions.

With these changes the following works:

```lua
config.action("revisions.toggle_select", function()
  jjui.builtin.revisions.toggle_select()
  jjui.builtin.revisions.move_down()
end)
```

With one caveat: the default`toggle_select` behaviour has always been to jump to parent. I removed that from the code and instead defined it in a backward-compatible way in `config.toml`

```toml
[[actions]]
name = "revisions.toggle_select"
lua = """
 jjui.builtin.revisions.toggle_select()
 jjui.builtin.revisions.jump_to_parent()
 """
```

I think this approach is very flexible and should enable a range of customisation paths.

### DIFF
--- a/internal/config/default/config.toml
+++ b/internal/config/default/config.toml
@@ -36,3 +36,11 @@ binding_profile = ":builtin"
 
 [ssh]
   hijack_askpass = false
+
+# Backwards compatibility
+[[actions]]
+name = "revisions.toggle_select"
+lua = """
+ jjui.builtin.revisions.toggle_select()
+ jjui.builtin.revisions.jump_to_parent()
+ """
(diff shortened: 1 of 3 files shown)


idursun:
I am merging this for now to get going with the upcoming changes.

PR #578, opened 2026-03-07
feat(help): add "help view" bound to `f1`

idursun:
This PR reintroduces the dedicated Help screen as a reimplementation of the previous view that was removed when we moved to expanded status help. Expanded status help remains for quick inline hints, and the full Help view is now back for complete keybinding discovery across scopes, with filtering and scrolling.

The Help view is currently opened with `ctrl+/`. I initially wanted to bind it to `shift+?`, but that is not feasible because producing `?` already requires shift on my keyboard, so `ctrl+/` is the most practical binding for now. I am open to suggestions.

image
image


baggiiiie:
love to see help menu back, tho not benefiting me much personally anymore but still a good functionality for jjui as a program to have

wondering if double question marks `??` could be used, `ctrl+/` feels a bit awkward and unintuitive

idursun:
> wondering if double question marks ?? could be used, ctrl+/ feels a bit awkward and unintuitive

I agree, we cannot do ?? though because first ? will be processed immediately as _expand status_. I think we can bind help to `f1` which makes more sense and discoverable than `ctrl+/`.

baggiiiie:
> we cannot do ?? though because first ? will be processed immediately as expand status

i was thinking, if status menu is expandable, `?` expands it, and a subsequent `?` triggers the help page

f1 works well too tho, either way it's good!

PR #584, opened 2026-03-11
[experiment] interactive bookmark view

baggiiiie:
As the interactive bookmark view has been highly requested, and there are quite a bit of discussions going around it, I'm experimenting with a half-page bookmark view. 

### bookmark operation

- Delete (`d`) - jj bookmark delete (batch-capable)
- Forget (`f`) - jj bookmark forget (batch-capable)
- Track (`t`) - jj bookmark track (batch-capable, auto-resolves default remote)
- Untrack (`u`) - jj bookmark untrack (batch-capable)
- Rename (`r`) - jj bookmark rename (opens input prompt)
- Create (`c`) - starts a create-bookmark operation (delegates to revisions pane)
- Move (`m`) - starts a move-bookmark operation (delegates to revisions pane)
- Push (`P`) - jj git push --bookmark (respects remote context)
- Fetch (`F`) - jj git fetch --branch (respects remote context)

### revision integration

- Show bookmark in revision view (`s` on a bookmark) - navigate the revisions pane to the bookmark's commit (if visible)
- Set revset (`alt+enter` on a bookmark) - set revision view's revset to `::$BookmarkName` (revset could be changed)
- Edit (`e`) - jj edit the bookmark's target revision (with confirmation)
- New (`n`) - jj new from the bookmark's target revision (with confirmation)

### others

- Select (`space`) - select multiple bookmarks to allow batch opertion, like create a new revision after selecting two bookmarks
- sorting - bookmarks sorted by: local present > local deleted > remote-only, then by proximity to current commit
- Expand/collapse (`enter`) - tree structure; top-level bookmark can expand to show its remote refs
- Filter by remote (`left/right/l/h`) - cycle through "all", "local", or specific remote names (clickable tabs)
- Text filter - fuzzy-matches bookmark name, target, commit ID, remote name
- Focus management (`tab`) - pane can be focused/unfocused, scope-aware keybindings
- Persistence across refreshes - preserves cursor position, expanded state, selections, and filter across data reloads
- Mouse support - click items, click remote tabs, scroll
- Ace jump support (`f`) during `c`reating and `m`oving bookmark

## screen recording

below screen recordings are made during an early stage of this development, keybindings have changed. 

### toggle, focus, and show details

https://github.com/user-attachments/assets/e320f936-abbe-4d78-b2e1-ec5548db829b

### rename, delete, forget, new

https://github.com/user-attachments/assets/87fe54ac-6afa-4e78-adee-8999d9fb0e6f

### go to revision from bookmark 

https://github.com/user-attachments/assets/1b91f98d-e9cf-42d4-b155-b6e9ce3c5b36

### filtering and move
 
https://github.com/user-attachments/assets/b11a4805-001e-4124-9c63-e8072a920950

### change revset in revision view

https://github.com/user-attachments/assets/04ef1465-9a81-4c61-b37b-4241f2ebb6c6

## notes

apologies for the long duration of this PR, it took me quite a while to nail down some UX, and iron out the implementations. life came along too and distracted me for a while from working on it

truly appreciated for everyone who has tried and provided feedback! it keeps me motivated!


idursun:
Hey thanks @baggiiiie !

I'll have a look and report back.

idursun:
I had a chance to play with it. I like that it shows as a panel on the right and I like moving in and out of revisions using tab. Very nice.

I also like that local and remote are visually distinguished from each other but I think it would be nice to show which remote the bookmark belongs to. I have 6 different remotes and cannot tell the remote of a bookmark until i hit space. We can show the name of the remote as `bookmark@remote` but this won't work for bookmarks that are repeated in multiple remotes (e.g. main)

I like that move is showing where the bookmark is going to be moved visually but this can lead to user errors as a remote can be moved to any revision. Current bookmark menu only lists moveable bookmarks to the selected revision and doesn't allow user to move any other if they are not in the same line of history. I think this is important but I don't have recommendation at the moment.

I'd like locals to be listed at the top of the list as they are the ones most likely that I will perform an action on. Even better if they are listed by distance as the current bookmark menu is doing so I would already be on a bookmark that I am going to move (as it would be the closest moveable bookmark).

AlejandroSanchez90:
looks good! one suggestion while you are there! select multiple bookmarks to create a merge!, rightnow im doing revset bookmark() selecting all bookmarks i want to merge and pressing 'n'

thanks for your work!!

baggiiiie:
hi all thanks for the feedback! truly appreciated all of them!


## feedback and solutions

> I think it would be nice to show which remote the bookmark belongs to. 

- showing remote name along with bookmarks
- for the case with 6 bookmarks, for now it should be hidden with hint (truncating might be better?)

> as a remote can be moved to any revision. Current bookmark menu only lists moveable bookmarks to the selected revision and doesn't allow user to move any other if they are not in the same line of history. 

- `move` operation on unmovable bookmarks are now shown with a flash message

> locals to be listed at the top ... listed by distance

- you have it, done!

> being able to select multiple bookmarks to create a merge

- a `select` could be implemented, and i'd like to keep consistent with revision view (`space`)
- keybinding would be: `space` as select, `enter` to open details, `alt+enter` to jump to revision, and `L` to set revset. 
	- note, `shift+enter` is considered but isn't captured by bubbletea. 
	- `L` to set revset as that's how we set revset in revision view right now

- fwiw, you can now `M` to set parent, and `tab` to go in bookmark view, go to the destination bookmark, and press `enter` to jump there and set the revision as parents
	- it's quite a handful to type down the workflow, but i think using it isn't that bad? try it for yourself and let me know!

## others

`t`rack and `u`ntrack to be implemented

image

nickchomey:
Could tab be used to cycle through the different remotes, as happens in the current bookmark menu? This would reduce clutter

baggiiiie:
hey all 
i've been dog fooding this feature for 2 weeks now, made some changes along the way:
- added `c`reate, git `P`ush and `F`etch
- makes certain interaction more logical and intuitive, like go to revision from bookmark
- keybinding changes from last update:
    - `enter` to open details of a bookmark
    - `alt+enter` to go to its revision (if visible in revision view)
    - `L` to set revset to `::$bookmark` (`L` is chosen as that's what we use in revision too) 
- code cleanup and bug fixes

thinking about adding:
- maybe a confirmation card for `f`orget, `d`elete, since they are easily triggered
- newly added `jj bookmark advance` support
- conflicted bookmark handling
- allow remote name as filter item, since i couldn't find a good UX with "tab to cycle remotes" mentioned by @nickchomey 

@idursun don't worry much about code yet, i'll do it final polishing once this feature meets expectation!

thanks to everyone's contribution!

nickchomey:
I finally got around to building this branch and trying it out.

1. I dont quite understand what the problem was with adding something like how the current bookmarks modal cycles through remotes with tab.

https://github.com/user-attachments/assets/c34377ab-6ac8-4286-8a5a-e32c23039097

I see that tab toggles between the log and bookmarks pane, and that's nice. But perhaps shift+tab or something similar (and user-configurable) could be used to cycle between bookmarks in this new view?

2. It would _definitely be essential_ to add a confirmation card for f, d, u, and anything that mutates things. I just used a few of those accidentally because i was focused on the bookmarks pane rather than the log pane (fortunately the op log is so simple to access and use)
3. It would be good if we could switch panes with the mouse as well - ive tried a few times to click the log while the bookmarks pane is focused, and it doesn't let me. I have to use tab, then click
4. I can't figure out what order the bookmarks are shown in. It changes all the time. I think i'd prefer alphabetical order by default, and perhaps there can be a toggle for sorting by the order in which they show up in the log? And then show some indication of which order is currently being used?

baggiiiie:
hey @nickchomey thanks for the feedback! truly appreciate it!

- for 1, tab is currently used for switching focus between revision panel and bookmark panel, not sure which key might be better; also, switching bookmarks doesn't play out quite well with the current display. im still experimenting, trying to come up with something i think is nice, before showing it here
- 2 and 3 are addressed!
- the bookmarks are displayed based on their distance to current revision, this was brought up by idursun [here](https://github.com/idursun/jjui/pull/584#issuecomment-4042459390), which i think is quite nice. let me know if it make more sense now that you know the underlying logic!

idursun:
@baggiiiie Sorry for the conflicts, I had ran it back then but didn't have a chance to provide feedback.

I like the idea behind this, and we can make this work. I don't know if you have any motivation left to rebase this on top of main. Totally fair if not. I can have a look at salvaging the work done here.

We can introduce a configuration option that replaces the bookmarks with this one and make this one the default once we reach the feature parity with the old bookmarks.

baggiiiie:
hey @idursun i just resolved the conflicts, don't worry about it! 

apologize on my end for dragging this PR for so long. i've been caught up with work and some other stuff, didn't have the time to give it one last polish 

will try to push iron it out this week and open reviews :D

nickchomey:
I get this error when I try to move a bookmark "sideways"

image

Should `--allow-backwards` be set by default, or somehow accessible via `alt+enter`?

Apart from this PR, this also happens if I just use `B` to set the bookmark, which calls `jj bookmark set -r ry asdf` and gives the same error. 

Also, interestingly, the current bookmark modal seems to filter out bookmarks that are not part of the selected branch

image

So, this interactive view is superior in the regard that it allows you to see and operate on all bookmarks

internal/jj/bookmark_action_commands.go
```diff
@@ -0,0 +1,36 @@
+package jj
```
baggiiiie, line 1: this file addresses the issue of bookmark state in bookmark pane, it decides whether an action is valid based on parsed bookmark state, then calls the lower-level constructor, which comes from `internal/jj/commands.go`.

i was thinking whether this should be merged into `internal/jj/commands.go`, decided to leave it as a single file as it contains more than calling commands.

internal/ui/bookmarkpane/tree.go
```diff
@@ -0,0 +1,321 @@
...
+	if a >= 0 && b >= 0 {
+		return a - b
+	}
+	if a < 0 && b < 0 {
+		return b - a
+	}
+	return b - a
+}
+
+func calcDistanceMap(current string, commitIDs []string) map[string]int {
```
baggiiiie, line 316: all this distance related logic is duplicated from `bookmarks.go`. we can make it shareable (or remove `bookmarks.go` if this new feature eventually wins the heart of everyone)

internal/ui/bookmarkpane/tree.go
```diff
@@ -0,0 +1,321 @@
+package bookmarkpane
```
baggiiiie, line 1: this file is probably doing too much: 
1. converts bookmarks into row nodes
1. builds visible tree rows
1. handles expansion/filtering
1. implements search matching
1. sorts bookmarks by local/remote status + distance

other than separating into different files logically, i don't have a better way to deal with this complexity

internal/ui/operations/bookmark/create_bookmark.go
```diff
@@ -0,0 +1,109 @@
+package bookmark
```
baggiiiie, line 1: this feature might not be necessary. 
when user presses `c` in bookmark pane to *c*reate a bookmark, it triggers a ui like squash/rebase in the revision view, where user can move cursor and create a bookmark on a revision

internal/ui/secondary_panes.go
```diff
@@ -0,0 +1,169 @@
+package ui
+
```
baggiiiie: this file manages preview pane and bookmark pane. it handles their split layout, visibility, focus switching, preview restoration after closing bookmarks, and etc

internal/config/default/bindings.toml
```diff
@@ -5,6 +5,8 @@ bindings = [
     { key = "?", action = "ui.expand_status", scope = "ui", desc = "expand status help" },
     { key = "f1", action = "ui.open_help", scope = "ui", desc = "help" },
     { key = "ctrl+z", action = "ui.suspend", scope = "ui", desc = "suspend" },
+    { key = "shift+x", action = "ui.toggle_bookmark_view", scope = "ui", desc = "bookmark view" },
```
baggiiiie: placeholder keybinding, to be updated..

internal/ui/bookmarkpane/actions.go
```diff
@@ -0,0 +1,319 @@
+package bookmarkpane
+
```
baggiiiie, line 2: actions allowed in bookmark pane

baggiiiie:
hey @idursun , apologies for the long duration of this PR, it took me quite a while to nail down some UX, and iron out the implementations. life came along too and distracted me for a while from working on it

i have tried out different designs, the code implementation is AI-assisted, with my careful reviews. i've updated the PR description to have a list of features and added a few comments to help with the review process of this 4000+ LOC diff.

a few more implementation details could be addressed, love to have your feedback 😄

internal/ui/common/msgs.go
```diff
@@ -50,9 +50,9 @@ type (
 		Ordered bool
 	}
 	ShowInputMsg struct {
-		Title  string
-		Prompt string
-		Value  string
+		Title        string
+		Prompt       string
+		InitialValue string
```
idursun: This change seems random and unrelated to this PR

internal/ui/input/input.go
idursun, line 1: I don't know why `input` is changed in this PR

internal/ui/operations/bookmark/set_bookmark.go
idursun, line 1: I prefer this operation is not changed and don't know anything about the bookmark pane. If we let this now, then we will end up creating coupling between various parts of the app. I also couldn't find how and where in the bookmark pane this operation is triggered, so it doesn't make sense for me now that it is trying to go focus back to the bookmark pane.

internal/ui/secondary_panes.go
idursun: I see. We don't have focused panels and this is an attempt to introduce that mechanism but it is very purpose built and even though the name sounds like a general concept, the implementation is highly coupled to the preview and bookmark pane. 

I personally would try delivering this missing building block first and then use it while building an interactive panel like bookmark pane.

internal/ui/ui.go
```diff
@@ -43,27 +45,32 @@ import (
...
+	frameCursor             *tea.Cursor
+	width                   int
+	height                  int
+	activeSplit             *split
+	previewSplit            *split
+	bookmarkSplit           *split
+	secondaryPaneActive     secondaryPaneKind
+	secondaryRestoreOnClose secondaryPaneKind
+	bookmarkPaneFocused     bool
+	bookmarkPane            *bookmarkpane.Model
```
idursun: This list will keep growing as we keep adding more views like this. This is obviously not fault of this PR but this PR is only adding more fuel to the fire.

idursun (commented):
Hey thanks!

I quickly skimmed through the code and also tested the new panel. It is good overall but it is also changing too much across the codebase. I left comments to the ones I noticed in the first pass.

I think majority of the changes are coming from the fact that we don't have a built-in panel management/focus tracking mechanism and this PR side stepping the issue by adding special more conditions across the codebase. It works, but in my eyes it is not maintainable by me. 

I also noticed that some of the keybindings are not aligned with the rest of the app. For example, `tab` used to change remotes but it doesn't because it is used for changing the focus, but then `l` or `right` key used to expand (look into details of an item) but it is changing the remote. I understand why they are made this way, but definitely tripped up my muscle memory.

I'll do another review pass when I have the time. 

Overall, I like the bookmark pane, but I don't like the fact that a lot of things had to change to make this work.

internal/ui/operations/bookmark/set_bookmark.go
baggiiiie: gotcha, updated and moved "restore focus to bookmark pane" into the root ui model, since the pane orchestration already lives there

internal/ui/secondary_panes.go
baggiiiie: hey thanks! this was my concern too. but didn't wanna increase the blast radius for this codebase. now the refactoring has come:

the `secondary_panes.go` file is now gone. instead a `internal/ui/split/` package is introduced that, which separates the concerns:

- `split.go`: a generic Split layout component that handles rendering any primary/secondary panels. handles separator dragging behavior. it has no knowledge of the panes, only takes two `ImmediateModel` values and a `SplitState`. could be extended to a list if we do have more panels.

- `panel.go`: a `SplitPanel` orchestrator. defines a `Panel` interface and manages which secondary pane, and pone focuses. 

- The root `internal/ui/split.go` is now thin wiring. delegates panel level intents to `SplitPanel` via `handlePanelIntent`, keeps the root `HandleIntent` clean.

- as a result, `preview.go` got cut down. it no longer owns layout positioning (which now lives in `SplitState`), as layout is a concern for `Split`, not for preview. `preview` model is left with its core responsibility: fetching and rendering content, and tracking its own visibility.

- the focus tracking (previously scattered across the codebase) now lives entirely within `SplitPanel.focusedPane`, synced outward via `syncFocus()`.

internal/ui/ui.go
baggiiiie: with the new introduction of a `SplitPane` to manage bookmark and preview, this is *a little* thinner

internal/ui/split/panel.go
baggiiiie, line 1: a `SplitPanel` orchestrator that sits on top of `Split`. defines a `Panel` interface and manages which secondary pane is active and where focus lives

internal/ui/split/split.go
baggiiiie, line 1: a generic `Split` layout component that handles rendering any primary/secondary pair with a draggable separator. It has no knowledge of what the panes are. only takes in two `ImmediateModel` values and a `SplitState`.

internal/ui/split/split.go
```diff
@@ -0,0 +1,264 @@
...
+		MaxPercent: 95,
+	}
+	s.clamp()
+	return s
+}
+
+func (s *SplitState) SetPosition(position config.PreviewPosition) {
+	s.AutoPosition = position == config.PreviewPositionAuto
+	s.AtBottom = position == config.PreviewPositionBottom
+}
```
baggiiiie: this is some side effect, not ideal: right now `Preview` and `BookmarkPane` are both split, they currently share the same config state. it's only a temporary implementation to seek feedback

@idursun do u think we should 
- allow `BookmarkPane` to have its own config for width percentage, position, and etc, 
- or share it with preview, and introduce a new "SplitPanelWidth" kind of config?

internal/ui/split.go
baggiiiie, line 1: only for wiring up preview, bookmark, and root UI now

internal/ui/preview/preview.go
baggiiiie, line 1: `preview.go` got simplified as some logics are now handled by `ui/split/split.go`

baggiiiie:
hey @idursun thanks for the review!

i pushed some changes (in a separate commit for easier codereview), they are not final, as now i have a decision to make regarding bookmark pane's config options like `width percentage`, which needs your input 

> some of the keybindings are not aligned with the rest of the app

hmm this, i haven't found a better solution. if `tab` is used to cycle remotes, and `l/h` to switch focus between bookmark pane and revisions, then it clashes with `l/h` in revision :(

internal/ui/split/split.go
idursun: Good question!

You know what, I'd like to see this split panel work in a PR of its own that's changing how the revisions + preview is rendered. That can be a good addition regardless and then we can clearly see what are the moving parts. (e.g. what should be the visual cue for a focused panel)

If we can have these split panels nested then `preview percentage` doesn't really make sense from that point on. Split panels are expected split at 50%. I am not sure if I am making sense though.

internal/ui/revisions/revisions.go
```diff
@@ -71,6 +71,7 @@ type Model struct {
 	ensureCursorView       bool
 	requestInFlight        bool
 	checkedRevisions       map[string]appContext.SelectedRevision
+	focused                bool
 }
```
baggiiiie, line 77: added for the focus-switching mechanism between revisions and bookmark pane

internal/ui/split/container.go
```diff
@@ -7,66 +7,124 @@ import (
...
-	OnShow()
+	common.ScopeProvider
+	OnShow() tea.Cmd
+	OnHide()
+	SetFocused(bool)
 }
 
-// SplitContainer manages one optional split-content pane next to a primary view.
+type overlayContent interface {
+	RenderOverlay(dl *render.DisplayContext, box layout.Box)
```
baggiiiie, line 20: bookmark pane as `SplitContent` needs a way to trigger confirmation rendering for operations like delete

internal/ui/split/container.go
```diff
@@ -7,66 +7,124 @@ import (
...
 
-// SplitContent is a renderable model that can be shown inside a SplitContainer.
+// SplitContent is content that can be displayed beside the primary view.
 type SplitContent interface {
 	common.ImmediateModel
-	OnShow()
+	common.ScopeProvider
+	OnShow() tea.Cmd
+	OnHide()
+	SetFocused(bool)
```
baggiiiie, line 16: as Bookmark pane is focus-able with its own keybinding, it needs a few lifecycle methods like OnShow to reset pane state and clear stale bookmark selections.
OnHide to deal with focus state, confirmation cleanup etc

internal/ui/split/container.go
```diff
@@ -7,66 +7,124 @@ import (
...
 
-func (sc *SplitContainer) Resize(delta float64) bool {
-	old := sc.state.Percent
+func (sc *SplitContainer) FocusSplitContent() {
+	sc.setContentFocused(true)
+}
+
+func (sc *SplitContainer) FocusPrimary() {
+	sc.setContentFocused(false)
+}
```
baggiiiie, line 116: Primary is the main view (revisions/oplog)

internal/ui/split/renderer.go
```diff
@@ -9,82 +9,47 @@ import (
...
-
 // SplitRenderer lays out two immediate-mode models with a draggable separator.
 // It handles the low-level rectangle splitting, separator drawing, and drag hitbox.
 type SplitRenderer struct {
-	State    *SplitState
-	Vertical bool
-	// Primary could be revision, oplog
-	Primary common.ImmediateModel
-	// Secondary could be preview, bookmark pane
-	Secondary          common.ImmediateModel
```
baggiiiie, line 22: `SplitRenderer` does not own UI models anymore, it only computes layout, separator, dragging box 
instead, `SplitContainer` owns content, focus, scopes, lifecycle

internal/ui/split/renderer.go
```diff
@@ -9,82 +9,47 @@ import (
...
-func (s *SplitRenderer) Init() tea.Cmd {
-	return nil
-}
-
-func (s *SplitRenderer) Update(msg tea.Msg) tea.Cmd {
-	return nil
-}
-
-func (s *SplitRenderer) ViewModels(dl *render.DisplayContext, box layout.Box, primary, secondary common.ImmediateModel, vertical bool) {
+func (s *SplitRenderer) Render(dl *render.DisplayContext, box layout.Box, primary, secondary common.ImmediateModel, vertical bool) {
```
baggiiiie, line 26: SplitRenderer now only receives the models for the current frame, only handles rendering logic

baggiiiie:
@idursun hey, i just rebased this PR to main, it should be ready for review now!

baggiiiie:
hey @idursun let me know if there's anything i can do to make code review easier for this big PR (e.g., more atomic commits, stacked PR, etc), or maybe put this behind a feature flag :D

idursun:
Oh sorry, I somehow missed the ping from this one.

> or maybe put this behind a feature flag

This is a good idea. Maybe we can add a configuration option `new_bookmarks_enabled = true` (or something like that) that would show this one when `b` is pressed.

I'd frame this as baggiiiie's experimental bookmarks view. I'd assign all the issues and discussions about the new view to you. 

I tried it and I am afraid it doesn't click for me. I still find the flow and keybindings very cumbersome, but it might just be me problem.

We can keep the experiment going until we have enough feedback to know if folks find it useful or not. Based on that we can retire the experiment.

Does that sound like a plan?

baggiiiie:
> Maybe we can add a configuration option new_bookmarks_enabled = true (or something like that) that would show this one when b is pressed.

@idursun for sure, sounds good! let's put it in so people can try it out and we can collect some feedback from discord etc to see where this feature should go! 

i'll work on another PR to add the config option first. do u think something like below is good (like in [open codex's config](https://learn.chatgpt.com/docs/config-file/config-basic#enabling-features))? imo this would be nice if we'd like to have more experiments like this in the future
```toml
[features]
interactive_bookmark_pane = true
```

or one config that's specific to bookmark, like 
```toml
[bookmark]
interactive_bookmark_pane = true
```

either way, we don't have existing options for them, another PR is needed for this

---

update:
went with the second option since its blast radius is small. created this PR https://github.com/idursun/jjui/pull/721

nickchomey:
I've been eager to test this out since you started 5 months ago, but never really did because it seemed like it was always sort of in draft/experiment mode. If it is somewhat "stable" now, i think an experimental flag would be great - i'd enable it and be happy to provide feedback

idursun (approved):
Thanks! Let's get this in and see where it goes. I'll treat this as experiment for the time being and let you shape it based on the feedback.

nickchomey:
Looking forward to testing this!

nickchomey:
Where would you like feedback? perhaps create a new issue or discussion?

PR #590, opened 2026-03-12
fix: Use wcwidth for final UI screen rendering

baggiiiie:
Bug: 
Certain emojis break the rendering of Preview border, causing the border
to be left shifted. 

image

Root cause:
the final Ultraviolet screen buffer used GraphemeWidth, which counts
variation-selector emoji like "⬇️" as 2 cells. Many terminals render
that glyph in 1 cell, so pane separators in the preview layout appeared
shifted when such a glyph was present in revision text. Regular emoji
like "😭" still measure as width 2, which is why they did not reproduce
the bug.

Fix:
render the final screen buffer with ansi.WcWidth instead, and add a
regression test documenting the differing widths for "⬇️" and "😭".

### DIFF
--- a/internal/ui/ui.go
+++ b/internal/ui/ui.go
@@ -84,6 +84,14 @@ const (
 	scopeCommandHistory   keybindings.Scope = "command_history"
 )
 
+// Use wcwidth-style measurement for the final screen buffer.
+//
+// This matches how terminals typically place variation-selector emoji such as
+// "⬇️": GraphemeWidth counts it as 2 cells, but terminals commonly render it
+// in 1 cell. Using WcWidth keeps pane separators aligned with terminal layout.
+// Regular emoji like "😭" still measure as width 2.
+var screenWidthMethod = ansi.WcWidth
+
 func (m *Model) Init() tea.Cmd {
 	return tea.Batch(m.revisions.Init(), m.scheduleAutoRefresh())
 }
@@ -366,7 +374,7 @@ func (m *Model) View() string {
 
 	box := layout.NewBox(layout.Rect(0, 0, m.width, m.height))
 	screenBuf := uv.NewScreenBuffer(m.width, m.height)
-	screenBuf.Method = ansi.GraphemeWidth
+	screenBuf.Method = screenWidthMethod
 
 	if m.diff != nil {
 		m.renderDiffLayout(box)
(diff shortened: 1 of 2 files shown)


baggiiiie:
I feel like this is terminal shenanigans. most likely to be a terminal-display alignment problem, not a logical layout bug


`ansi.WcWidth` works on mine (ghostty) perfectly for ⬇️. tested on stock MacOS terminal too.


not sure if this is a universal solution, but the fact that ⬇️ breaks and 😭 doesn’t is a grapheme/wcwidth mismatch

idursun:
I had changed it to use `GraphemeWidth` in https://github.com/idursun/jjui/commit/c3ac691dfd2a3cef430eea78ff04df66ac14e813 3 days ago. Emojis render fine on ghostty for me. Misalignment happens on kitty though.

Also, `ansi.WcWidth` is the default of screen buffer, so this is effectively reverting the change I made. I think we need to set the width function based on the capabilities of the underlying terminal. Some can work fine with graphemes, and some don't.

baggiiiie:
hmm what's the reason for changing it to `GraphemeWidth` ? 


> I think we need to set the width function based on the capabilities of the underlying terminal

i think functions like `lipgloss.Width(...)` and maybe `ansi.StringWidth(...)` is assuming rune/grapheme-based segment width, which might not necessarily agree with ansi.WcWidth.

Even with this fix, it still has some leftover work to be done. after the revision being highlighted, the border disappeared, see below 


i'll have a look into this another time, feel free to close this PR or include this change as you see fit

https://github.com/user-attachments/assets/7f335261-eead-45f3-bb65-7ce43f35b14c

idursun:
> Even with this fix, it still has some leftover work to be done. after the revision being highlighted, the border disappeared, see below

Yes, so lipgloss.Width is using grapheme width and Bubbletea v2 is enabling grapheme clustering if the terminal supports it. ScreenBuffer defaulted to WcWidth and that's why I changed it to be aligned with the terminal's default capability. It solved the problem for Ghostty and friends. However, Kitty is defaulting to WcWidth and that's why this is not working fine there. 

I vaguely remember that Bubbletea v2 release notes (or some other discussion) that there's a message that we can listen to to know which width method the terminal supports. I'll try implementing that first and will add new functions to the `render` package so that we use one source of truth across the board.

idursun:
I am closing this in favour of https://github.com/idursun/jjui/pull/592

PR #592, opened 2026-03-13
fix(render): detect terminal width method

idursun:
This fixes misaligned lines in terminals that use grapheme clustering by default, such as Ghostty, while preserving correct rendering in terminals like Kitty that do not.

We detect terminal support for Unicode core width mode by listening for `tea.ModeReportMsg`, then we store the detected width method into the `render` package. I moved `Width` methods to the render package so that it is consistent across the codebase (by using the detected width method).

I tested it on both terminals and seems to be rendering correctly.


baggiiiie (commented):
nice! didnt expect this will require such big change

PR #621, opened 2026-03-29
fix(textinput): handle shift+backspace in terminals with Kitty keyboard protocol

matze:
Disclaimer: I am not a Go developer and this was mostly figured out by Claude Code.

Terminals like wezterm and ghostty on Linux report shift+backspace as a distinct key event via the Kitty keyboard protocol. The bubbles textinput default KeyMap only binds plain backspace, so shift+backspace was silently ignored. Add a PatchTextInputKeys helper that extends the KeyMap to also accept shift+backspace, shift+delete, and shift+alt+backspace, and apply it to all textinput instances.

Fixes #619

### DIFF
--- a/internal/ui/common/autocompletion/autocompletion.go
+++ b/internal/ui/common/autocompletion/autocompletion.go
@@ -62,7 +62,7 @@ func WithCompletionsDisabled() Option {
 }
 
 func New(provider CompletionProvider, options ...Option) *AutoCompletionInput {
-	ti := textinput.New()
+	ti := common.TextInputNew()
 	ti.Focus()
 	ti.Prompt = ""
 	ti.ShowSuggestions = true
--- a/internal/ui/common/textinput.go
+++ b/internal/ui/common/textinput.go
@@ -1,0 +1,25 @@
+package common
+
+import "charm.land/bubbles/v2/textinput"
+
+// Create a textinput.Model extending the KeyMap so that shift+backspace and
+// shift+delete behave the same as their unshifted counterparts. Some terminals
+// (e.g. wezterm, ghostty on Linux) report these as distinct key events via the
+// Kitty keyboard protocol, but the bubbles default KeyMap does not include
+// them.
+func TextInputNew() textinput.Model {
+	ti := textinput.New()
+
+	km := &ti.KeyMap
+	km.DeleteCharacterBackward.SetKeys(
+		append(km.DeleteCharacterBackward.Keys(), "shift+backspace")...,
+	)
+	km.DeleteCharacterForward.SetKeys(
+		append(km.DeleteCharacterForward.Keys(), "shift+delete")...,
+	)
+	km.DeleteWordBackward.SetKeys(
+		append(km.DeleteWordBackward.Keys(), "alt+shift+backspace")...,
+	)
+
+	return ti;
+}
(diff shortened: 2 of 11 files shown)


internal/ui/common/autocompletion/autocompletion.go
```diff
@@ -63,6 +63,7 @@ func WithCompletionsDisabled() Option {
 
 func New(provider CompletionProvider, options ...Option) *AutoCompletionInput {
 	ti := textinput.New()
+	common.PatchTextInputKeys(&ti)
```
baggiiiie: if we were to introduce this change, instead of a "patch", i think it's better to have something like `common.TextInputNew()` that does what `PathTextInputKeys` is doing *and* `textinput.New()`
it should be able to remove this two-step (textinput.New then PatchTextInput) boilerplate

otherwise every caller of textinput.New needs to call the patch

internal/ui/common/textinput.go
```diff
@@ -0,0 +1,20 @@
...
+	km := &t.KeyMap
+	km.DeleteCharacterBackward.SetKeys(
+		append(km.DeleteCharacterBackward.Keys(), "shift+backspace")...,
+	)
+	km.DeleteCharacterForward.SetKeys(
+		append(km.DeleteCharacterForward.Keys(), "shift+delete")...,
+	)
+	km.DeleteWordBackward.SetKeys(
+		append(km.DeleteWordBackward.Keys(), "shift+alt+backspace")...,
+	)
```
baggiiiie, line 22: intuitively, i think a `DeleteWorkForward` is missing?

internal/ui/common/textinput.go
```diff
@@ -0,0 +1,20 @@
...
+func PatchTextInputKeys(t *textinput.Model) {
+	km := &t.KeyMap
+	km.DeleteCharacterBackward.SetKeys(
+		append(km.DeleteCharacterBackward.Keys(), "shift+backspace")...,
+	)
+	km.DeleteCharacterForward.SetKeys(
+		append(km.DeleteCharacterForward.Keys(), "shift+delete")...,
+	)
+	km.DeleteWordBackward.SetKeys(
+		append(km.DeleteWordBackward.Keys(), "shift+alt+backspace")...,
```
baggiiiie: see [tea package - charm.land/bubbletea/v2 - Go Packages](https://pkg.go.dev/charm.land/bubbletea/v2#Key.Keystroke): 

> Note that modifier keys are always printed in the following order:
>
> - ctrl
> - alt
> - shift
> - meta
> - hyper
> - super
> 
> For example, you'll always see "ctrl+shift+alt+a" and never "shift+ctrl+alt+a".


i think it should be `alt+shift+backspace`

baggiiiie (changes requested):
hey @matze thanks for contributing, a few issues i think worth addressing here

internal/ui/common/textinput.go
matze: Changed it and it also works, yes.

internal/ui/common/autocompletion/autocompletion.go
matze: Changed to your suggestion and now without AI :-)

internal/ui/common/textinput.go
matze: But … with which key?

idursun:
Hey sorry for the late reply, but shouldn't this contribution be sent to the downstream https://github.com/charmbracelet/bubbles repository instead?

We don't own textarea/textinput and I personally don't think we should be patching downstream issues here.

idursun:
Seems abandoned

PR #627, opened 2026-04-13
fix(details): sync details selection for routed intents

baggiiiie:
## Bug
- Right now, when preview is open, moving selected files in details doesn't update preview.

## Cause
- `details` kept the selected file in sync when updates flowed through `Update()`, but routed intents can bypass that wrapper and call `HandleIntent()` directly. 
- That meant keyboard navigation in the details overlay could move the cursor without updating `MainContext.SelectedItem`.
- selection behavior such as preview refreshes and context-aware file commands still pointed at the previously selected file even though the highlighted row had moved.

## Fix
- added a single `updateSelection()` helper in the details
- updated the file to use it where the current file can change:
	- after loading the file list, 
	- after click handling, 
	- after cursor changes in `Update()`,
	- after handled intents in `HandleIntent()`.

### DIFF
--- a/internal/ui/operations/details/details.go
+++ b/internal/ui/operations/details/details.go
@@ -105,26 +105,13 @@ func (s *Operation) Update(msg tea.Msg) tea.Cmd {
 		}
 		s.setItems(items)
 
-		// Set selection to current cursor position
-		var selectionChangedCmd tea.Cmd
-		if current := s.current(); current != nil {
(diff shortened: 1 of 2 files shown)


idursun (approved):
Geez, thanks!

Our current update selection approach is a huge foot gun. I don't have a better solution at the moment, so I guess we will keep patching these issues until something better comes up.

PR #628, opened 2026-04-13
fix(set_parents): leak set_parents operation key scope

baggiiiie:
Right now, when set_parents is triggered, no keys are leaked are key strokes like `j` `k` doesn't work

Following the pattern in for rebase operation, add leakAll to set_parent's scope


idursun (approved):
Thanks!

PR #629, opened 2026-04-13
fix: add back esc to clear selected revisions

baggiiiie:


baggiiiie:
sending another one 😆

baggiiiie:
@idursun do u think it's a good idea to have some e2e style testing for jjui? i don't have much opinions on this, probably good to spot behavior regressions. 

i've been using agents to use tmux for verifying TUI behaviors. maybe we can have a set of bash scripts consisting of `tmux send-keys`, or something like how the gifs are generated for jjui-recording? 

these two are just on top of my head, maybe there're some existing mature framework to do this for TUI

idursun:
Hey thanks!

I could swear that I tested this!

> do u think it's a good idea to have some e2e style testing for jjui? i don't have much opinions on this, probably good to spot behavior regressions.

We have a fair amount of features and I've been manually testing what I usually use but I keep missing some behaviour that I don't use very often. e2e tests makes sense, but I also don't know how to do this.

> i've been using agents to use tmux for verifying TUI behaviors. maybe we can have a set of bash scripts consisting of tmux send-keys, or something like how the gifs are generated for jjui-recording?

This is interesting, I'll do some research to see what's out there.

idursun (approved):

PR #639, opened 2026-04-17
feat(revisions): stream quick search until match or revset exhausted

### LINKED ISSUE #638: Quick search silently misses matches in unloaded rows
## Description

`/` quick search only matches rows that have already been streamed into the in-memory buffer. On broad revsets (e.g. `all()`) in large repos, most of history hasn't been fetched yet, so `/foo` can silently miss matches that do exist — giving the impression the repo doesn't contain them. There's no indication the search was partial.

`internal/ui/revisions/revisions.go`'s `search()` runs `common.CircularSearch` over `m.rows` only; `m.hasMore` and `requestMoreRows` are never consulted from the search path.

1. Open jjui on a large repo with a broad revset: `jjui -r 'all()'`.
2. Let the initial viewport load (a few rows).
3. Press `/` and search for a string that only appears in old history.

## Expected Behavior

`/` searches the full evaluated revset — streaming more rows as needed until a match is found or the revset is exhausted.

## Actual Behavior

Cursor stays put. No "not found" feedback. The match was real, just in rows that hadn't been streamed yet.

## Specifications

- Platform: macOS (not platform-specific)
- Version: main @ d4e573c

proever:
## Summary

- `/` and `n`/`N` now keep pulling rows from the `GraphStreamer` until a match is found or the revset is exhausted, instead of only searching what's already in the in-memory buffer.
- `common.CircularSearch` now returns `(int, bool)` so callers can distinguish "no match" from "match-at-cursor"; the `oplog` caller discards the bool (oplog has no streaming, so behavior is unchanged).
- A new `pendingQuickSearch` field on the revisions model records an unresolved search; `appendRowsBatchMsg` re-runs the search as each batch lands and requests more rows while `hasMore` stays true. Cleared on `RevisionsQuickSearchClear` and on a fresh `streamingReadyMsg`.

Closes #638

## Test plan

- [x] `go test ./...` passes
- [x] New tests in `internal/ui/revisions/revisions_quicksearch_test.go`: streams-until-match, stops-when-exhausted, doesn't-arm-when-hasMore-false, clear-cancels-pending
- [x] Updated `internal/ui/common/search_test.go` for the new return signature
- [x] Manual: `jjui -r 'all()'` on a large repo, `/` for an old string — cursor now lands on the match

### DIFF
--- a/internal/ui/common/search.go
+++ b/internal/ui/common/search.go
@@ -7,10 +7,10 @@ import (
 )
 
 // CircularSearch performs a circular search through searchable items.
-// Returns the index of the first matching item, or cursor if no match is found.
-func CircularSearch(items []screen.Searchable, query string, startIndex, cursor int, backward bool) int {
+// Returns (matchedIndex, true) when an item matches, or (cursor, false) otherwise.
+func CircularSearch(items []screen.Searchable, query string, startIndex, cursor int, backward bool) (int, bool) {
 	if query == "" {
-		return cursor
+		return cursor, false
 	}
 
 	n := len(items)
@@ -22,10 +22,10 @@ func CircularSearch(items []screen.Searchable, query string, startIndex, cursor 
 			c = (startIndex - i + n) % n
 		}
 		if matchesQuery(items[c], query) {
-			return c
+			return c, true
 		}
 	}
-	return cursor
+	return cursor, false
 }
 
 func matchesQuery(item screen.Searchable, query string) bool {
--- a/internal/ui/common/search_test.go
+++ b/internal/ui/common/search_test.go
@@ -49,6 +49,7 @@ func TestCircularSearch(t *testing.T) {
 		cursor     int
 		backward   bool
 		want       int
+		wantFound  bool
 	}{
 		{
 			name:       "empty query returns cursor",
@@ -56,6 +57,7 @@ func TestCircularSearch(t *testing.T) {
 			startIndex: 0,
 			cursor:     2,
 			want:       2,
+			wantFound:  false,
 		},
 		{
 			name:       "forward match from start",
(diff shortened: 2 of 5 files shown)


idursun:
Thanks for the PR.

As you have described this PR is loading all the revisions to find a matching text which would work fine on small repositories but would render the quick search useless on large repositories (the one I work against at work) as it would take a good 10s of seconds to load all of the revisions, which is kind of unexpected and not something I want to happen under the hood.

baggiiiie:
i think the idea of this feature has its own merit, maybe we can load revisions in a progressive way but i don't think the implementation will be easy (wiring up log view updates with quick search)

tho one thing to point out is, the use case of quick search is to quickly jump to a revision in the current `jj log` view. if u wanna search the whole log history, maybe use something like `jj log -r 'description(substring:"YOUR_STRING")'`?

proever:
Thanks for taking a look, and I understand the objections about large revsets. Would you consider a version that prompts the user to search another X revs?

I personally find the current behavior of `/` quite unintuitive in that I often use to it jump to revs that I *know* are there, but possibly just off screen. I actually initially thought I was just using it wrong.

wrs:
FWIW, I agree the current behavior is unintuitive and honestly thought the function was just broken. If I can see the commit, I can just ace-jump to it, I figured `/` was for exactly the case of jumping further back in the history without having to page-down.

A couple of suggestions, ignoring implementation questions since I'm just a happy user with dreams…

While streaming in the log, have some kind of `searching...` indicator if it takes longer than a second, and allow ^C to stop the search if you realize you went off the end (imitating `less` behavior).

Make a convenient search syntax that gets converted to a revset using `description()` etc. and string patterns. Then make some command like `L` but with that syntax. In other words treat search as a fliter on revisions rather than a movement command.

PR #640, opened 2026-04-18
feat(absorb): add target picker for jj absorb --into

### LINKED ISSUE #634: FR: support `jj absorb --into` / `--to`

jjui supports jj absorb, but with the default behavior, it absorbs from the selected revision into the default destination set. This means I can trigger absorb, but I cannot constrain or select which ancestor revisions are allowed as destinations.

This can get tricky with stacked commits where a later revision contains follow up fixes that should be absorbed only into a specific subset of earlier commits. The default mutable() destination set can be broader than I want, and I have no way in jjui to limit it.

It would be great if jjui supported the equivalent of:

```bash
jj absorb --into <revsets>
```

or

```bash
jj absorb --to <revsets>
```

so that absorb destinations can be constrained from within the UI.

Right now, the alternative is to drop into terminal and handle that outside jjui

Current absorb support does not expose destination selection.

## Additional context

This feature would be useful for people maintaining stacked branches or doing commit cleanup before submission.

A typical example is:
* revset A: backend work
* revset B: UI work
* revset C: follow-up fixes touching both

In this case, absorb can move the backend fix back into A and the UI fix back into B. Having support for target/destination selection would make it possible to constrain absorption to those specific revsets

manusajith:
#### Changes: 

shift+a runs jj absorb directly against jj's default mutable() destination set. Users with stacked changes need to constrain which ancestors absorb is allowed to target, which forced them out of jjui into the terminal.

- Convert absorb into a picker operation (sibling of set_parents/abandon)
- Pre-select jj's default --into candidates with << into >> markers
- space toggles a candidate, enter applies, esc cancels
- Toggled-off defaults render dimmed << default >> so they remain visible
- Unchanged target set runs plain jj absorb --from (current behavior)
- Net-modified set runs jj absorb --from <src> --into <id> per kept target
- Empty target set explicitly closes without running

closes #634

<hr>

#### Notes                                    
                                        
A few small things worth noting:
                                                                                   
For the default `--into` options, I've opted for `mutable() & ::source` rather than the plain `mutable()` you suggested. 
jj's CLI does use plain `mutable()` by default, but its absorb algorithm only ever assigns hunks to ancestors of the source. 
From jj's docs: "Only ancestors of the source revision will be considered." 
Narrowing it to `mutable() & ::source` keeps the picker from listing revisions that absorb would never target anyway.
Let me know if this works, or if you'd rather switch to plain `mutable()`.

I also had to rename the `revisions.absorb` action to `revisions.open_absorb`. The new operation needs a `revisions.absorb` scope, and that would collide with the existing `jjui.revisions.absorb()` function in Lua (the scope table would overwrite it). The new name also lines up with `open_abandon`, `open_squash`, etc.

https://github.com/user-attachments/assets/01fb9fd1-a919-4679-9fcb-16c8b309f517


idursun (approved):
Thanks, this is good!

> From jj's docs: "Only ancestors of the source revision will be considered."
Narrowing it to mutable() & ::source keeps the picker from listing revisions that absorb would never target anyway.
Let me know if this works, or if you'd rather switch to plain mutable().

I noticed this when I ran `jj log -r mutable()` which resolved to more revisions than the command normally operates on. I assumed command's help is not matching up the internal behaviour but didn't do any further investigation. What you have done makes sense.

As a stretch goal, I would convert absorb invocation from the details view to be like how squash is initiated but I can pick it up myself after this is merged.

PR #641, opened 2026-04-20
Dynamic light/dark switching

zerowidth:
This PR adds support for dynamic color theme notifications so the color theme can change if the system switches between light and dark modes.

- The tea UI now registers for `ansi.SetLightDark`, which enables CSI 2031 mode reporting (https://contour-terminal.org/vt-extensions/color-palette-update-notifications/)
- The UI responds to palette changes by emitting a `ThemeChangedMsg`, invalidating the currently-loaded palette's cache.
- A few components have their own local caches of the current palette styles, either by design (`Revision`, `Flash`) or implicitly by wrapping bubble tea components (`Autocompletion`, `Revset`) or both (`Status`). These are now refreshed explicitly after receiving a `ThemeChangedMsg`.
- Not all terminals support the color palette update notifications. Alacritty, which is what the zed embedded terminal uses, has explicitly chosen to not support these control codes. To support these terminal implementations, I've added a once-a-second "get current color theme" poll to the tea event loop. This hooks into the same `ThemeChangedMsg` notifications, just via polling instead of eager notification.
- At init, tea now explicitly requests whether CSI 2031 is supported. If it is supported, the polling loop is disabled.
- The `DEBUG` env var in the contributing doc was incorrect, I fixed it to refer to the correct `JJUI_DEBUG` var name.

I've tested this in iTerm2 (supports CSI 2031) successfully, as well as zed terminal (only supports polling).

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -83,7 +83,7 @@ This produces `catalog_gen.go` and `builtins_gen.go`. A staleness test will fail
 - Configuration handling is in the `internal/config/` directory
 - Default configuration and key bindings are in the `internal/config/default` directory
 - Operations (rebase, squash, split, etc.) are implemented as operations the `internal/ui/operations/` directory
-- Set `DEBUG=1` environment variable for printing debug messages to `debug.log` file
+- Set `JJUI_DEBUG=1` environment variable for printing debug messages to `debug.log` file
 
 ## Contributing Documentation
 
--- a/cmd/jjui/main.go
+++ b/cmd/jjui/main.go
@@ -222,37 +222,13 @@ func run() int {
 		appContext.TerminalThemeDetected = true
 	}
 
-	var defaultThemeName string
-	if appContext.TerminalHasDarkBackground {
-		defaultThemeName = "default_dark"
-	} else {
-		defaultThemeName = "default_light"
-	}
-
-	theme, err = config.LoadEmbeddedTheme(defaultThemeName)
+	theme, err = config.ResolveTheme(appContext.TerminalHasDarkBackground, appContext.JJConfig.GetApplicableColors())
 	if err != nil {
-		fmt.Fprintf(os.Stderr, "Error loading default theme '%s': %v\n", defaultThemeName, err)
+		fmt.Fprintf(os.Stderr, "Error loading theme: %v\n", err)
 		return 1
 	}
 
-	var userThemeName string
-	if appContext.TerminalHasDarkBackground {
-		userThemeName = config.Current.UI.Theme.Dark
-	} else {
-		userThemeName = config.Current.UI.Theme.Light
-	}
-
-	if userThemeName != "" {
-		theme, err = config.LoadTheme(userThemeName, theme)
-		if err != nil {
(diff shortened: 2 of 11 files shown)


baggiiiie (dismissed):
nice! 

i pulled down the code and played around, found that it doesn't work for op_log, maybe some cache issue too

zerowidth:
> i pulled down the code and played around, found that it doesn't work for op_log, maybe some cache issue too

Interesting, there's definitely more I missed. If I go into `oplog` and switch themes before returning to the main UI, it never updates that either. I'll keep digging.

idursun:
Hey @zerowidth thanks for working on this.

I need some time to study the code. I skimmed the diff and approach makes sense. This can potentially be extended to changing theme at run time as well.

zerowidth:
I _think_ I've covered everything (merge conflict aside), via entirely mechanical means. Copy, paste, add test... 

I don't like this very much at all.

I wonder if there's a different approach we can use to get the colors updated in a single place? Except the lipgloss elements take `lipgloss.Style` as arguments, rather than being able to do lazy lookups. Lazy lookups on the custom elements might simplify some of it, though?

idursun:
Hey, thanks.

> I don't like this very much at all.

This is painful. The current mechanics of theme support is not the ideal way, it is just how things evolved over time. This is one of the examples where it would trigger a change-the-world type of refactoring spree for me.

> I wonder if there's a different approach we can use to get the colors updated in a single place?

So we have the palette and it caches styles after they are resolved, but we also have styles fields in the models which also caches styles. Given that we are doing immediate mode rendering, we render every frame from scratch, which means we can *technically* move all the styles into the ViewRect methods. This means if the palette changes what we render changes. I'd deliver this as a foundational refactoring PR. Then I'd follow it up with ThemeChangedMsg changes which only invalidates the palette cache, so we can have dynamic theme change. The last PR would be the one that's polling the terminal for the light/dark change.

Does this sound like a plan to you?

idursun:
@zerowidth @baggiiiie I'll do the refactor of moving cached styles from models to render methods so that we can salvage the rest of the work done in this PR.

idursun:
I pushed my changes to https://github.com/idursun/jjui/tree/inline-theme-styles-into-renderers

It's moving style resolution into viewrect methods and seems to be working fine. I had some local test code which updates the palette and loads a random theme from `themes` folder and I was able to see them change during run time. If that's ok I can commit that branch and then rebase this branch on top of that.

zerowidth:
I've now rebased _just_ the core pieces (palette cache invalidation, notify subscription) of this work on `main` so it's a clean PR again. If you run the branch from #651 along with this one, I think we have light/dark sorted.

zerowidth:
I noticed in `iterm2` that after I quit or suspended `jjui`, i'd get terminal garbage in my term when changing the color mode. This was because `jjui` had subscribed to the mode 2031 notifications, but didn't unsubscribe on exit/suspend. That should be fixed now!

idursun (approved):
Hey thanks! 

I will merge the linked PR first and then merge this.

https://github.com/user-attachments/assets/e7137a1b-5850-4f7e-9000-dff8eb653b65

PR #642, opened 2026-04-20
feat: include directories in path entries for fuzzy_files

baggiiiie:
Rename files→paths and add directory entries (e.g. "src/", "src/pkg/")
to the fuzzy file search so users can filter by directory. The
buildPathEntries helper scans each path for '/' separators and adds
every ancestor prefix before the file itself, deduplicating with a
seen set.

Also simplify the PreviewScroll construction to a direct type
conversion, and group the CommandArgs/CommandWithStdin type block.

### DIFF
--- a/internal/jj/commands.go
+++ b/internal/jj/commands.go
@@ -29,11 +29,13 @@ const (
 	JJUIPrefix                  = "_PREFIX:"
 )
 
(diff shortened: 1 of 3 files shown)


idursun (approved):
Thanks!

PR #643, opened 2026-04-20
fix(ui): unable to dismiss flash message after split command error

baggiiiie:
Bug: pressing esc in details after a split command error did not dismiss the flash message.

Why it happened:
- In details scope, esc resolves to revisions.details.cancel, which maps to DetailsClose.
- That close intent was handled before the top-level ui.cancel path that dismisses flash messages.
- As a result, esc closed details (or was otherwise consumed by a close/cancel handler) instead of dismissing the visible flash.

Fix:
- Give flash dismissal priority for cancel/close-style intents in the top-level UI.
- When a flash is visible, dismiss it before routing cancel/close intents to details, oplog, help, choose, command history, or other close operations.
- Keep the original close behavior on the next esc once the flash is gone.

UX improvement:
- Dismiss flash before close operations, so transient errors are cleared first and esc behaves more predictably across views.

Bug detail:

https://github.com/user-attachments/assets/9e8a633a-132c-487e-a77a-7eb65f645ffa

### DIFF
--- a/internal/ui/operations/details/details.go
+++ b/internal/ui/operations/details/details.go
@@ -30,12 +30,13 @@ type updateCommitStatusMsg struct {
 }
 
 var (
-	_ operations.Operation         = (*Operation)(nil)
-	_ operations.EmbeddedOperation = (*Operation)(nil)
-	_ common.Focusable             = (*Operation)(nil)
-	_ common.Editable              = (*Operation)(nil)
-	_ common.Overlay               = (*Operation)(nil)
-	_ dispatch.ScopeProvider       = (*Operation)(nil)
+	_ operations.Operation                = (*Operation)(nil)
+	_ operations.EmbeddedOperation        = (*Operation)(nil)
+	_ operations.CheckedItemsSynchronizer = (*Operation)(nil)
+	_ common.Focusable                    = (*Operation)(nil)
+	_ common.Editable                     = (*Operation)(nil)
+	_ common.Overlay                      = (*Operation)(nil)
+	_ dispatch.ScopeProvider              = (*Operation)(nil)
 )
 
 type Operation struct {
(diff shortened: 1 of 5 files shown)


idursun (changes requested):
Hey thanks, 

Bug is real but I don't think this is the fix. 

I don't want to special case flash dismissal like this. I have done a ton of clean up just to get rid of special casings across the codebase.

The bug is in [internal/ui/revisions/revisions.go:580](https://github.com/idursun/jjui/blob/f136b9c82bcb221f42ec6dff1f81cc163684232b/internal/ui/revisions/revisions.go#L578)

```go
		// Clear checked items and reset to default operation
		if len(m.context.CheckedItems) > 0 || !m.InNormalMode() {
			m.context.ClearCheckedItems(reflect.TypeFor[appContext.SelectedRevision]())
			m.resetOperations()
			return nil, true
		}
```

The bug doesn't happen when no file is selected. It only happens when one or more files are selected using space. When the command fails, checked items are left behind and the condition holds which effectively prevents `intent.Cancel` reaching to ui.HandleIntent.

I'd try to fix this by clearing the checked items instead of patching flash dismissal.

baggiiiie:
hey @idursun thanks for the review, finally have some time to revisit this

im getting a bit disconnected with the newer codebase now, didn't have a good look at the original change i opened :( 

now updated to clear SelectedFile checked items whenever the base operation leaves the details view. also, previously, when flash message and details both open, `esc` close details first, now changed the behavior such that `esc` closes flash message first (more intuitive imo)

what do you think?

idursun (approved):
Hey thanks!

I'd personally just fix the bug by counting the selected items of Revision type instead of all the selected items, but this will do too.

I see that we have a new interface defined but it is only used in one place (which is a smell) but I can clean it up later on because whole selection tracking needs a rethink.

baggiiiie:
@idursun thanks! i will be more careful !

PR #647, opened 2026-04-22
Add `golangci-lint` in CI

nikosavola:
Add `golangci-lint` in CI. Maybe the tool is a bit opinionated but would surely be useful with a right config.

### DIFF
--- a/.github/workflows/go.yml
+++ b/.github/workflows/go.yml
@@ -19,6 +19,10 @@ jobs:
           go fmt ./...
           go mod tidy
           git diff --exit-code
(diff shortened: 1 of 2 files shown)


idursun:
Hey thanks for the PR but I think I am going to pass this one. 

I don't really like CI failing even though everything on my machine is working fine and tests are passing. I occasionally get burnt by not running `go fmt` prior to pushing which is on me but I don't want to add more checks unless it is absolutely necessary. I don't think go-lint is necessary at this stage of the project.

PR #651, opened 2026-04-26
refactor: inline theme styles into renderers

idursun:
We store the resolved styles in the `CurrentPalette` instance but also models had a reference to `styles` instances during runtime. This PR moves style resolution to ViewRect functions so that we only rely on the palette cache, effectively allowing theme change during runtime just by updating the palette.

Prerequisite for #641


zerowidth:
Nice! Hopefully moving the style resolution to a more central location makes #641 easier. I rebased my branch on top of this, and ignoring all the mechanical changes and focusing on just the light/dark change notifications worked!

PR #652, opened 2026-04-27
feat(bookmark): prefill bookmark name via open_set_bookmark value arg

rdeaton:
Adds an optional `value` arg to `revisions.open_set_bookmark` so Lua scripts can prefill the bookmark input dynamically. Also introduces `$string?(name)` annotation syntax for optional string args (mirroring `$bool()`).

Example `config.lua` — prefills with `<user>/<date>/` so you only type the suffix:

```lua
function setup(config)
  -- Drop the default shift+b binding so we can replace it.
  for i = #config.bindings, 1, -1 do
    local b = config.bindings[i]
    if b.action == "revisions.open_set_bookmark" and b.scope == "revisions" then
      table.remove(config.bindings, i)
    end
  end

  config.action("set_bookmark_prefixed", function()
    local prefix = "rdeaton/" .. os.date("%Y%m%d") .. "/"
    jjui.revisions.open_set_bookmark({ value = prefix })
  end, { scope = "revisions", key = "shift+b", desc = "set bookmark" })
end
```

### DIFF
--- a/cmd/genactions/main.go
+++ b/cmd/genactions/main.go
@@ -809,10 +809,10 @@ func validateSetValueType(fieldType, value string, enums map[string][]enumValueM
 		if strings.HasPrefix(value, "\"") && strings.HasSuffix(value, "\"") {
 			return nil
 		}
-		if isStringArg(value) {
+		if isStringArg(value) || isOptionalStringArg(value) {
 			return nil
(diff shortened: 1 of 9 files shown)


idursun (approved):
Thanks!

PR #663, opened 2026-05-05
feat(ui): config setting to completely disable mouse support

### LINKED ISSUE #661: FR: config setting to completely disable mouse support
i prefer the jjui to not catch any mouse events, to have only standard terminal mouse events processing so i can just select and copy any text.

Would be nice to have a setting to disable jjui mouse support completely. 

shift+click is ok to select and copy, but i do not need any mouse support in jjui at all, so i would feel more comfortable knowing that my mouse clicks will not do anything in jjui.

shoce:
Add NoMouse bool `toml:"no_mouse"` to UIConfig to allow disabling mouse with `no_mouse = true` in config ui section. 

fixes #661

### DIFF
--- a/internal/config/config.go
+++ b/internal/config/config.go
@@ -110,8 +110,9 @@ type UIConfig struct {
 	SetWindowTitle bool             `toml:"set_window_title"`
 	// TODO(ilyagr): It might make sense to rename this to `auto_refresh_period` to match `--period` option
 	// once we have a mechanism to deprecate the old name softly.
-	AutoRefreshInterval        int `toml:"auto_refresh_interval"`
-	FlashMessageDisplaySeconds int `toml:"flash_message_display_seconds"`
+	AutoRefreshInterval        int  `toml:"auto_refresh_interval"`
+	FlashMessageDisplaySeconds int  `toml:"flash_message_display_seconds"`
+	NoMouse                    bool `toml:"no_mouse"`
 }
 
 func GetExpiringFlashMessageTimeout(c *Config) time.Duration {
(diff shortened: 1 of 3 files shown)


idursun:
Hey, thanks.

Instead of using double negative, can we name it as `mouse_support` and set the default value to true in the config.toml?

shoce:
> Instead of using double negative, can we name it as `mouse_support` and set the default value to true in the config.toml?

sure, pushed a new commit.

shoce:
Add MouseSupport bool toml:"mouse_support" to UIConfig to allow disabling mouse with `mouse_support = false` in config `[ui]` section.

internal/config/config.go
```diff
@@ -110,8 +110,25 @@ type UIConfig struct {
...
+			if pBool, isBool := p.(bool); isBool {
+				u.MouseSupport = pBool
+			} else {
+				return fmt.Errorf("invalid type for 'mouse_support' in ui configuration: expected bool, got %T", p)
+			}
+		} else {
+			u.MouseSupport = true
+		}
+	}
+	return nil
```
idursun: You don't need this because you can simply add the default value to the [config.toml](https://github.com/idursun/jjui/blob/main/internal/config/default/config.toml)

idursun (dismissed):
I have one comment, I can merge this and fix it myself, or if you want to keep going then feel free to do so. Decision is up to you.

internal/config/config.go
shoce: thanks, removed it!

idursun (approved):
Thanks!

PR #667, opened 2026-05-07
feat(ui): add ctrl-n/p list navigation aliases

### LINKED ISSUE #666: FR: ctrl-p / ctrl-n for cycling up / down through list items
Most applications (most notably emacs and vim) support <kbd>ctrl</kbd><kbd>p</kbd> and <kbd>ctrl</kbd><kbd>n</kbd> for cycling through lists, basically as aliases to the <kbd>↑</kbd> and <kbd>↓</kbd> cursor keys. It's convenient to not have fingertips leave the home row. 

Is it possible to add those two keybindings as aliases for the two cursor keys to allow that behaviour?

vvvu:
## What changed

This PR adds `ctrl+n` / `ctrl+p` as cosy home-row aliases for moving down and up in input-driven lists.

This matches the behaviour many folks already expect from readline, Vim-ish and Emacs-ish interfaces, and makes these little lists nicer to use without reaching for the arrow keys.

Closes #666.

## Why

`j` and `k` are already lovely in the main revision list, but they are not a good fit inside text inputs because they should remain normal typed characters there.

`ctrl+n` and `ctrl+p` give those input lists a familiar navigation path without stealing regular typing.

## Notes

While adding the revset aliases, I also adjusted the inactive revset scope ordering so preview scrolling keeps its existing `ctrl+n` / `ctrl+p` behaviour when the revset input is not being edited.

## Verification

- `go build ./cmd/jjui`
- `go test ./...`

### DIFF
--- a/internal/config/default/bindings.toml
+++ b/internal/config/default/bindings.toml
@@ -23,8 +23,8 @@ bindings = [
     { key = "enter", action = "revset.apply", scope = "revset", desc = "apply" },
     { key = "tab", action = "revset.autocomplete", scope = "revset", desc = "autocomplete" },
     { key = "shift+tab", action = "revset.autocomplete_back", scope = "revset", desc = "autocomplete back" },
-    { key = "up", action = "revset.move_up", scope = "revset", desc = "up" },
-    { key = "down", action = "revset.move_down", scope = "revset", desc = "down" },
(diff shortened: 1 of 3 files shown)


idursun (approved):
Thanks! 

Nice catch about the revset + preview ordering issue. I'll revisit that area at some point.

PR #670, opened 2026-05-10
fix: Use config preview commands for diff

imericxu:
**Bug**: Pressing `d` to access the full-page diff doesn’t show the same content as the preview (i.e., the configured command)

**Steps to reproduce the issue (if applicable)**:
1. Set a custom preview command in the config
2. Notice the preview and the actual diff are different (this is especially noticeable if using Delta)

**How this fixes the problem**: Use the configured `RevisionCommand` and `FileCommand` for the `d` (diff) action.

> [!NOTE]
> I didn’t add tests because I’m not super familiar with Go.

### DIFF
--- a/internal/ui/operations/details/details.go
+++ b/internal/ui/operations/details/details.go
@@ -12,6 +12,7 @@ import (
 	"charm.land/bubbles/v2/key"
 	tea "charm.land/bubbletea/v2"
 	"charm.land/lipgloss/v2"
+	"github.com/idursun/jjui/internal/config"
 	"github.com/idursun/jjui/internal/jj"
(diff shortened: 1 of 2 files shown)


idursun:
Hey, thanks.

This is working as intended, not a bug. Preview is user configurable and shows what the user wants to show for the selected item.

The `d` action always opens the diff view. It does not use the preview command, and it is not expected to match preview output.

Also, I recommend opening an issue before creating a PR. I’m going to close this for now, but feel free to open an issue if you think I missed something.

imericxu:
@idursun I’ve created an issue for this at #671 .

PR #675, opened 2026-05-13
fix(exec): parse jj prompt input with shellwords (#655)

### LINKED ISSUE #655: `:` prompt splits on whitespace, breaking quoted arguments
Hi! Love jjui, thanks for building it. This isn't a big deal because `$` works fine, but the fix is easy, so it's probably worth considering.

Quoted arguments typed at the `:` prompt are mis-tokenized because the input is split on whitespace with no quote handling. Press `:`, type `new -m 'add skills'`, and jj reports:

```
Error: Failed to parse revset: Syntax error
1 | skills'
  |       ^---
```

The pipeline from input string to jj argv has no shell-aware splitter. `ExecMsgFromLine` only trims whitespace:

https://github.com/idursun/jjui/blob/96911e3f14b377fdb2d7c89e510c7cd558ee9098/internal/ui/exec_process/exec_process.go#L18-L31

and the `ExecJJ` branch of `ExecLine` splits with `strings.Fields`:

https://github.com/idursun/jjui/blob/96911e3f14b377fdb2d7c89e510c7cd558ee9098/internal/ui/exec_process/exec_process.go#L37-L39

So `new -m 'add skills'` becomes argv `[new, -m, 'add, skills']`. jj's `-m` consumes `'add` as the message and `skills'` is left as a positional revset — which matches the trailing single quote in the error. Reproducible at the shell:

```
$ jj new -m "'add" "skills'"
Error: Failed to parse revset: Syntax error
1 | skills'
```

Probably went unnoticed because most `:` uses don't need quoted args (`new`, `abandon`, `rebase -d X`, `git fetch`, `bookmark create foo`), inline describe is a separate path, and anyone who hit it once likely fell back to `$` shell mode, which works because it hands the line to `$SHELL -c`.

Fix is small — swap `strings.Fields` for a shell-style splitter like [`mattn/go-shellwords`](https://pkg.go.dev/github.com/mattn/go-shellwords) on the `ExecJJ` branch only. The `ExecShell` branch is unaffected. On parse error (e.g., unbalanced quotes), the existing handler in `internal/ui/status/status.go` already re-opens the prompt with the typed line preserved, which is the right behavior. Happy to send a PR if that sounds good.

I tested it locally and it seems to work well.

```diff
diff --git a/internal/ui/exec_process/exec_process.go b/internal/ui/exec_process/exec_process.go
index 7503882d75..f5ab529836 100644
--- a/internal/ui/exec_process/exec_process.go
+++ b/internal/ui/exec_process/exec_process.go
@@ -2,6 +2,7 @@
 
 import (
 	"bufio"
+	"fmt"
 	"io"
 	"os"
 	"os/exec"
@@ -10,6 +11,7 @@
 
 	tea "charm.land/bubbletea/v2"
 	"github.com/charmbracelet/x/term"
+	shellwords "github.com/mattn/go-shellwords"
 	"github.com/idursun/jjui/internal/jj"
 	"github.com/idursun/jjui/internal/ui/common"
 	"github.com/idursun/jjui/internal/ui/context"
@@ -35,7 +37,15 @@
 	replacements := ctx.CreateReplacements()
 	switch msg.Mode {
 	case common.ExecJJ:
-		args := strings.Fields(msg.Line)
+		args, err := shellwords.Parse(msg.Line)
+		if err != nil {
+			return func() tea.Msg {
+				return common.ExecProcessCompletedMsg{
+					Err: fmt.Errorf("parsing command line: %w", err),
+					Msg: msg,
+				}
+			}
+		}
 		args = jj.TemplatedArgs(args, replacements)
 		return execProgram("jj", args, ctx.Location, nil, msg)
 	case common.ExecShell:
```

SAY-5:
Fixes #655. The `:` prompt tokenized input with `strings.Fields`, so `new -m "add skills"` reached jj as `new -m "add skills"` (two args) and the trailing quoted half hit jjs revset parser. Swap to `mattn/go-shellwords` on the `ExecJJ` branch only; unbalanced quotes surface as `ExecProcessCompletedMsg.Err`, which the existing handler in `internal/ui/status/status.go` already re-opens the prompt for.

Per @idursun in https://github.com/idursun/jjui/issues/655#issuecomment-4347190209 ("Your suggestion makes sense and feel free to send a PR"). Added unit tests for plain-args parity with `strings.Fields`, quoted-arg grouping, and the error path. `nix/vendor-hash` will need a refresh on a machine with `nix` (run `nix run .#update-vendor-hash`), I do not have it locally.


idursun (approved):
Thanks!

PR #687, opened 2026-06-03
feat: add Home/End key support for navigating to top/bottom of revs

nikosavola:
Simply adds Home and End key support for navigating to top and bottom of revs, respectively.

### DIFF
--- a/internal/ui/intents/revisions_intents.go
+++ b/internal/ui/intents/revisions_intents.go
@@ -90,6 +90,16 @@ const (
 	TargetWorkingCopy
 )
 
+//jjui:bind scope=revisions action=go_to_top
+type GoToTop struct{}
+
+func (GoToTop) isIntent() {}
+
+//jjui:bind scope=revisions action=go_to_bottom
+type GoToBottom struct{}
+
+func (GoToBottom) isIntent() {}
+
 //jjui:bind scope=revisions action=move_up set=Delta:-1
 //jjui:bind scope=revisions action=move_down set=Delta:1
 //jjui:bind scope=revisions action=page_up set=Delta:-1,IsPage:true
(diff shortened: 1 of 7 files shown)


internal/ui/intents/revisions_intents.go
```diff
@@ -88,12 +88,16 @@ const (
 	TargetParent
 	TargetChild
 	TargetWorkingCopy
+	TargetTop
+	TargetBottom
```
idursun: I wouldn't add these here because the other ones are sort of relative to a revision but the new ones are just UI related so that turns the NavigationTarget into a mix bag of values.

internal/ui/revisions/revisions.go
```diff
@@ -908,6 +908,14 @@ func (m *Model) navigate(intent intents.Navigate) tea.Cmd {
...
 		m.ensureCursorView = ensureView
 		return m.updateSelection()
+	case intents.TargetTop:
+		m.SetCursor(0)
+		m.ensureCursorView = ensureView
+		return m.updateSelection()
+	case intents.TargetBottom:
+		m.SetCursor(len(m.rows) - 1)
+		m.ensureCursorView = ensureView
+		return m.updateSelection()
```
idursun: This is technically correct, but would only work for the loaded batch. If graph had more revisions then user has to press `j` and then `end` again to keep going. 

Instead I'd do the following so that it would update the selection but also would request more rows to be loaded so that the user can keep pressing `end` to jump through batch. I think that would be the expected behaviour.

```go
		return tea.Sequence(m.updateSelection(), m.requestMoreRows(m.tag.Load()))
```

idursun (changes requested):
Hey thanks! and sorry for the late review.

There are two comments that I thought worth flagging. 

I am happy to fix them if you don't feel like it but wanted to give you the opportunity first. Let me know what you think!

nikosavola:
Hey @idursun, asking for re-review

idursun (approved):
Thanks!

PR #690, opened 2026-06-06
fix: set filesearch list half viewport height

baggiiiie:
As I was working on the refactoring for preview and bookmarkpane, I found out that file search is emitting a message to update preview

https://github.com/idursun/jjui/blob/281650c7e24389d97a68561b96d0b885e7279983/internal/ui/ui.go?plain=1#L590-L597

Didn't know this feature exists at all... right now it's useless because the file_search list expands to the whole screen and covers whatever's underneath. 

Therefore I'm opening this change to make file_search list half height of the viewport. 

| before | after |
| -- | -- | 
| image | image |


idursun (approved):
Thanks!

idursun:
I am going to retire this fuzzy files soon though. It's preview controls are not working which I fixed locally but overall the whole experience is clunky. I think first class files view with a file tree and showing the annotations is worth exploring.

PR #698, opened 2026-06-19
fix(flash): prevent command output from corrupting card borders

baggiiiie:
certain commands like Git and SSH errors can contain carriage returns
(`\r`). When rendered directly inside a flash card, these characters
move the terminal cursor to the start of the line, causing subsequent
output to overwrite the card's left border.

Normalize CRLF and standalone carriage returns to newlines before
rendering flash messages and command history entries. Add regression
tests covering both error and history output.

## before 
file-5dd330798229614dbbf387723eb7f1b8

## after 
file-29e6190fdb1ba70724ab35998ca4ccee

### DIFF
--- a/internal/ui/flash/render.go
+++ b/internal/ui/flash/render.go
@@ -10,8 +10,7 @@ import (
 
 const commandMarkWidth = 3
 
-type CardRenderer struct {
-}
+type CardRenderer struct{}
 
 func NewCardRenderer() CardRenderer {
 	return CardRenderer{}
@@ -54,6 +53,7 @@ func (r CardRenderer) renderCard(command, text string, commandErr error, maxWidt
 			bodyText = commandErr.Error()
 		}
 		if bodyText != "" {
+			bodyText = render.ReplayTerminalOutput(bodyText)
 			bodyStyle := lipgloss.NewStyle()
 			if highlight {
 				bodyStyle = statusStyle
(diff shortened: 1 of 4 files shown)


idursun (dismissed):
Thanks!

baggiiiie:
@idursun thanks for the quick review, the previous fix somehow duplicates of `git:`, im looking into why

image

baggiiiie:
seems like jj use `/r` for in-place line updates for git commands, source https://github.com/jj-vcs/jj/blob/760330afb9475494e1dd5841eb7d84153e4bfbea/lib/src/git_subprocess.rs#L807-L859 (mainly triaged by gpt55, i reviewed the process)

even with TERM=dumb it still has such behavior. `--quiet` removes it, but also removes the useful output. from below screenshot we can see, if we simply replace `/r` it'd cause such duplication of `git:`. maybe this issue deserves to be upstreamed to support non-tty usage of jj 

image

the way around this is to draw the content in an isolated screen buffer, which is the change in the new commit i just pushed 

image

internal/ui/ui.go
```diff
@@ -387,9 +387,8 @@ func (m *Model) View() string {
 	}
 
 	m.displayContext.Render(screenBuf)
-	finalView := screenBuf.Render()
 	m.frameCursor = m.displayContext.Cursor()
-	return strings.ReplaceAll(finalView, "\r", "")
+	return screenBuf.Render()
```
idursun: How does this relate to the changes made in the flash renderer? Seems out of place to me.

baggiiiie: i think the global `\r` stripping became redundant after this PR?

it's probably good to still keep a safe guard, i'll revert this change

idursun (approved):
Thanks!

Issue #701, opened 2026-06-25
FR: allow multiple bookmarks for jj git push


neil-sriv:
## Is your feature request related to a problem? Please describe.
I often want to push multiple bookmarks when creating or updating stacks of changes.

## Describe the solution you'd like
I'd like to be able to select multiple revs with bookmarks and press g, b to suggest the option of jj git push --bookmark <bookmark 1> --bookmark <bookmark 2> etc.

## Describe alternatives you've considered
I do not want to use jj git push --all because I have many historical branches that are not relevant. I do not want to use the inferred bookmarks from just jj git push since I often have megamerge, worktrees, or other async workers editing revs that my head is not currently associated with.

## Why current features fail to cover this request
Selecting revs does not seem to keep track of the bookmarks at those revs.

## Additional context
I've implemented this locally and it seems very simple - can make a PR if desirable.

idursun:
@neil-sriv can we consider this issue resolved?

neil-sriv:
Yes!

PR #702, opened 2026-06-25
feat(git): add push option for selected bookmarks in multi-select

neil-sriv:
## Summary

- When multiple revisions with bookmarks are selected and the git menu is opened (`g`), add a new push option that runs `jj git push --remote <remote> --bookmark <name> ...` for only the bookmarks on the selected revisions
- Deduplicate bookmark names across selected revisions and skip conflicted bookmarks
- Bind the new option to shortcut `b` in the push category

## Test plan

- [x] `go test ./internal/ui/git/`
- [x] Select multiple revisions with bookmarks, press `g`, filter push options, and confirm the aggregated bookmark push item appears
- [x] Run the new option and verify only the selected bookmarks are pushed to the chosen remote

### DIFF
--- a/internal/ui/git/git.go
+++ b/internal/ui/git/git.go
@@ -639,7 +639,8 @@ func (m *Model) createMenuItems() []item {
 			if b.Conflict {
 				continue
 			}
-			if !seenBookmarks[b.Name] {
+			pushable := b.Local != nil && (len(b.Remotes) == 0 || b.HasRemote(selectedRemote))
+			if pushable && !seenBookmarks[b.Name] {
 				seenBookmarks[b.Name] = true
 				selectedBookmarks = append(selectedBookmarks, b.Name)
 			}
--- a/internal/ui/git/git.go
+++ b/internal/ui/git/git.go
@@ -631,11 +631,17 @@ func (m *Model) createMenuItems() []item {
 		selectedRemote = ""
 	}
 
+	var selectedBookmarks []string
+	seenBookmarks := make(map[string]bool)
 	for _, commit := range revisions.Revisions {
 		bookmarks := loadBookmarks(m.context, commit.GetChangeId())
 		for _, b := range bookmarks {
 			if b.Conflict {
 				continue
+			}
+			if !seenBookmarks[b.Name] {
+				seenBookmarks[b.Name] = true
+				selectedBookmarks = append(selectedBookmarks, b.Name)
 			}
 			for _, remote := range b.Remotes {
 				items = append(items, item{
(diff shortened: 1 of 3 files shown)


internal/ui/git/git.go
```diff
@@ -677,6 +683,22 @@ func (m *Model) createMenuItems() []item {
 				command:  jj.GitPush(flags...),
 				key:      "c",
 			})
+
+		if len(selectedBookmarks) > 0 {
+			var bookmarkFlags []string
+			for _, name := range selectedBookmarks {
+				bookmarkFlags = append(bookmarkFlags, "--bookmark", name)
+			}
+			flags := append([]string{"--remote", selectedRemote}, bookmarkFlags...)
```
idursun, line 693: If the selected bookmarks have different remotes, then this command would push all of them to the currently selected bookmark. This needs to be generated for the bookmarks that match the current selected bookmark.

internal/ui/git/git_test.go
```diff
@@ -91,6 +91,27 @@ func Test_PushChange(t *testing.T) {
...
+	commandRunner.Expect(jj.GitPush("--remote", "", "--bookmark", "feature-a", "--bookmark", "feature-b"))
+	defer commandRunner.Verify()
+
+	selected := jj.NewSelectedRevisions(&jj.Commit{ChangeId: changeId1}, &jj.Commit{ChangeId: changeId2})
+	op := NewModel(test.NewTestContext(commandRunner), selected)
+	test.SimulateModel(op, op.Init())
+	_ = test.RenderImmediate(op, 100, 40)
+
+	test.SimulateModel(op, func() tea.Msg { return intents.GitOpenFilter{} })
+	test.SimulateModel(op, test.Type("git push --bookmark feature-a --bookmark feature-b"))
```
idursun: I don't think this usage is correct here.

The following is a better usage.

```go
	test.SimulateModel(op, intents.Invoke(intents.GitFilter{Kind: intents.GitFilterPush}))
	test.SimulateModel(op, intents.Invoke(intents.GitApplyShortcut{Key: "b"}))
```

internal/ui/git/git.go
```diff
@@ -631,12 +631,18 @@ func (m *Model) createMenuItems() []item {
...
 	for _, commit := range revisions.Revisions {
 		bookmarks := loadBookmarks(m.context, commit.GetChangeId())
 		for _, b := range bookmarks {
 			if b.Conflict {
 				continue
 			}
+			if !seenBookmarks[b.Name] {
+				seenBookmarks[b.Name] = true
+				selectedBookmarks = append(selectedBookmarks, b.Name)
+			}
```
idursun: `BookmarkList(... -a -r ...)` can return remote-only bookmarks (`b.Local == nil`), but `jj git push --bookmark <name>` expects a local bookmark name. Should this skip bookmarks where `b.Local == nil` before adding them to `selectedBookmarks`?

idursun (changes requested):
Even though this is in draft I thought I could leave a quick review.

neil-sriv:
Appreciate the review! Had left this in draft till I got a response on https://github.com/idursun/jjui/issues/701 but happy to move forward with it.

idursun (approved):
Hey, thanks for addressing the review feedback. The remote filtering, remote-only bookmark handling, and test interaction changes all look good.

One remaining edge case: `HasRemote` currently ignores the remote bookmark’s `Tracked` field. An untracked bookmark on the selected remote can therefore be included in the batch, causing `jj git push` to reject the entire command. I'll fix that separately after merging.

PR #703, opened 2026-06-29
Show revset history before completions

baggiiiie:
right now when jjui is freshly launched, typing `L` to set revset shows functions and aliases. after resvet is set once, `L` only shows history. 

personally i think it makes sense to always show everything, with history, if exists, shown at the top of the list


idursun (approved):
Thanks!

PR #714, opened 2026-07-21
docs: require jj 0.37 or later

baggiiiie:
Updated the minimum supported `jj` version from v0.36 to v0.37 across the documentation.

`jjui` uses the [`change_offset` keyword](https://github.com/idursun/jjui/blob/107dc4b1c4710c34650805941614cdeeb28a0508/internal/jj/commands.go?plain=1#L58), which was introduced in jj v0.37. A clean v0.36 installation fails to render the revision log.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -49,7 +49,7 @@ go run ./cmd/genactions
 ## Requirements
 
 - Go 1.24.2+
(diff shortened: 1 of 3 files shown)


idursun (approved):
Thanks!

PR #716, opened 2026-07-22
Allow pushing all bookmarks in selected changes

### LINKED ISSUE #715: `b` shortcut to push bookmarks is missing when only one change is selected
The `b` shortcut that pushes the bookmarks on the selected change(s) only appears in the Git menu when *multiple* changes are selected.

Select a single change and the item never shows up, even when that change has pushable bookmarks. The aggregation logic in `git.go` (the `if len(selectedBookmarks) > 0` block) is nested inside the `if hasMultipleRevisions` block, so it's unreachable for a single selection.

Steps to reproduce:
1. Select a single change that has a local bookmark.
2. Open the Git push menu.
3. The `b` "push selected bookmarks" item is absent.

nikosavola:
### What is the current behavior?
The `b` shortcut aggregates the pushable bookmarks on the selected change(s) and pushes them in a single command. The block that builds it was nested inside the `if hasMultipleRevisions` check, so it was unreachable whenever a single change was selected, regardless of how many bookmarks that change had.

### What is the new behavior?
`b` is available whenever the selected change(s) have at least one pushable bookmark, independent of how many changes are selected.

### Changes
- Moved the `if len(selectedBookmarks) > 0` block out of the `hasMultipleRevisions` branch in `internal/ui/git/git.go`.
- Reworded the item label to `Push all bookmarks in selected changes (%s)`.

Closes #715

### DIFF
--- a/internal/ui/git/git.go
+++ b/internal/ui/git/git.go
@@ -680,22 +680,22 @@ func (m *Model) createMenuItems() []item {
 				command:  jj.GitPush(flags...),
 				key:      "c",
 			})
+	}
 
-		if len(selectedBookmarks) > 0 {
-			var bookmarkFlags []string
-			for _, name := range selectedBookmarks {
-				bookmarkFlags = append(bookmarkFlags, "--bookmark", name)
-			}
-			flags := append([]string{"--remote", selectedRemote}, bookmarkFlags...)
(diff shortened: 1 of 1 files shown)


idursun (approved):
Thanks!

PR #720, opened 2026-07-31
Support confirmation for changing immutable commits

nikosavola:
Right now only edit has a way to retry against an immutable commit (the alt+e force-edit binding). Everything else (split, describe, squash, rebase, ...) just fails with no way to recover short of leaving the TUI.

This adds a generic fix: when any jj command fails with jj's "commit is immutable" error, jjui pops up a confirmation dialog offering to retry the same command with --ignore-immutable.

  - jj.IsImmutableError detects the failure (and specifically excludes the root-commit error, which the flag can't bypass anyway)
  - CommandCompletedMsg grows a lazy Retry field; shared helpers build the retry command for both the buffered and interactive (pty)
  command paths
  - new confirmation dialog (internal/ui/immutable), defaults to "No" since it can pop up unattended, one-shot guarded so a held key can't
  fire it twice


nickchomey:
This would be a significant improvement

idursun:
Hey thanks

> Everything else (split, describe, squash, rebase, ...) just fails with no way to recover short of leaving the TUI.

What do you mean here? We have alt+enter which adss this flag automatically.

nikosavola:
> Hey thanks
> 
> > Everything else (split, describe, squash, rebase, ...) just fails with no way to recover short of leaving the TUI.
> 
> What do you mean here? We have alt+enter which adss this flag automatically.

I think my alt+enter has been bound to a shortcut to open a terminal split and I missed this...

This PR offers a different way, I guess they could coexist but I understand if that seems unnecessary.

PR #721, opened 2026-08-03
feat(config): add [bookmark] config for bookmark pane

baggiiiie:
Add a BookmarkConfig struct and [bookmark] TOML section to prepare for gating the experimental interactive bookmark pane behind a config flag.

The interactive_bookmark_pane option defaults to false so existing users keep the classic bookmark menu. This is a no-op config addition.


idursun (approved):
Thanks! I am going to approve and merge this. I'd like it to be `ui.interactive_bookmark_pane` but it is not important. We can change it later if it makes more sense to folks.

PR #722, opened 2026-08-04
fix(parser): keep gutter one segment per rune when padding blank graph lines

shuvamk:
Parsing the shipped fixture `test/testdata/single-line-with-description.log` (a commit followed by a blank `│` graph line, as `builtin_log_comfortable` emits) and extending the row drops two gutter columns:

```
rows[0].Extend()  actual "│"   (1 segment)
                  expected "│  " (3 segments, row.Indent == 3)
```

Downstream, that shifts every embedded operation panel (details file list, inline describe, evolog, rebase preview) two columns to the left of the row's own text under a comfortable template.

### Cause

`GraphRowLine.chop` first splits the gutter into one segment per rune, then pads short lines up to the row indent. The padding step re-fattened the **last** segment's `.Text` with `strings.Repeat(" ", indent)`, undoing the per-rune split. `Row.Extend()` derives its mask from the gutter's segment count, so a blank line reporting 1 segment instead of 3 loses the trailing columns.

### Fix

Pad by appending one single-space `screen.Segment` per remaining indent column, each carrying the last gutter segment's `Style`, so the one-rune-per-segment invariant holds.

### Test

`TestParser_Parse_ExtendBlankGraphLine` in `test/log_parser_test.go` extends the fixture row and asserts the gutter is `"│  "`. I verified it fails without the source change (`actual "│"` vs `expected "│  "`) and passes with it.

`go test ./...` is green.

### DIFF
--- a/internal/parser/row_line.go
+++ b/internal/parser/row_line.go
@@ -99,8 +99,13 @@ func (gr *GraphRowLine) chop(indent int) {
 
 	// Pad with spaces if indent is not fully consumed
 	if indent > 0 && len(gr.Gutter.Segments) > 0 {
-		lastSegment := gr.Gutter.Segments[len(gr.Gutter.Segments)-1]
-		lastSegment.Text += strings.Repeat(" ", indent)
+		style := gr.Gutter.Segments[len(gr.Gutter.Segments)-1].Style
+		for range indent {
+			gr.Gutter.Segments = append(gr.Gutter.Segments, &screen.Segment{
+				Text:  " ",
+				Style: style,
+			})
+		}
 	}
 }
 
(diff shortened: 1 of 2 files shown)


baggiiiie (approved):
lgtm, thanks! 
btw for such UI fixes, it'd be better to include screenshots to show what the issue was and the fixed UI

baggiiiie:
merging this, repro and results:

## before
https://github.com/user-attachments/assets/1b723c44-2607-467b-84c9-f63f2b5141b9 

## after 
https://github.com/user-attachments/assets/1f00186d-f834-4688-af8b-f77ddcd642fd

PR #732, opened 2026-08-25
toggle bookmarkpane with `b` when config flag is set

baggiiiie:
as #584 and #721 have been merged, jjui users can now try out the experimental bookmark pane feature behind a config flag

in previous PR, it was still `X` to trigger bookmark pane, which is fairly awkward. this PR updates it to intercept `b` so it replaces the current bookmark stack UI

tho this implies that these two features are mutually exclusive (in my own use case i think it's fine, since im mainly using bookmarkpane now, but of course, i am biased!), i understand this might not fit into the vision of jjui 

@idursun feel free to merge this PR as you see fit, or close it if it doesn't :")


idursun (approved):
Thanks!

PR #734, opened 2026-08-27
fix(dispatch): treat Ctrl-[ as Escape

### LINKED ISSUE #710: Ctrl-[ does not behave as an alias for Escape
## Description

`Ctrl-[` does not behave as an alias for Escape in jjui.

I normally use `Ctrl-[` instead of the physical Escape key. In terminal applications, these are traditionally equivalent, and tools such as Vim, Neovim, and lazygit accept both for canceling or moving back.

In jjui, the physical Escape key works as expected, but `Ctrl-[` does not trigger the same cancel or back action.

jjui’s default bindings use `esc`, and the key dispatcher appears to compare configured bindings directly against Bubble Tea’s key string or keystroke representation. With modern terminal input handling, `Ctrl-[` may be reported separately as `ctrl+[` rather than normalized to `esc`.

A possible fix would be to treat `ctrl+[` as an alias for `esc` in the central key-matching function. That would cover existing Escape bindings and cancellation of unfinished key sequences without duplicating bindings in every scope.

1. Run jjui inside tmux.
2. Open a dialog, view, search, or operation that can be closed or canceled with Escape.
3. Press `Ctrl-[`.
4. Press the physical Escape key for comparison.

## Expected Behavior

`Ctrl-[` should perform the same action as the physical Escape key, including:

* Closing dialogs and overlays
* Canceling operations
* Clearing searches or selections
* Returning to the previous view
* Canceling an unfinished multi-key sequence

## Actual Behavior

Pressing `Ctrl-[` does not trigger the configured Escape action.

Pressing the physical Escape key works correctly.

## Specifications

* Platform: NixOS 26.05 (Yarara), Linux 6.18.36 x86_64, tmux 3.7b, `TERM=tmux-256color`
* Version: jjui 0.10.8

stfulldev:
## summary

- treat `Ctrl-[` as an alias for `Escape` in the central key matcher
- reuse existing `esc` bindings without duplicating keys across scopes
- make `Ctrl-[` cancel pending key sequences like physical Escape

Closes #710

## testing

- `go test ./internal/ui/dispatch`
- `go test ./...`
- `go build ./cmd/jjui`
- manually verified help rebase andd quick search cancelation using an
  enhanced-terminal `Ctrl-[` key event

### DIFF
--- a/internal/ui/dispatch/dispatcher.go
+++ b/internal/ui/dispatch/dispatcher.go
@@ -221,5 +221,9 @@ func isPrefix(full []string, prefix []tea.Key) bool {
 }
 
 func keyMatches(candidate string, key tea.Key) bool {
-	return candidate == key.String() || candidate == key.Keystroke()
+	keystroke := key.Keystroke()
+	if candidate == key.String() || candidate == keystroke {
+		return true
+	}
+	return candidate == "esc" && keystroke == "ctrl+["
 }
--- a/internal/ui/dispatch/dispatcher_test.go
+++ b/internal/ui/dispatch/dispatcher_test.go
@@ -80,17 +80,44 @@ func TestDispatcher_SwallowsMismatchedContinuationAndResets(t *testing.T) {
 	require.True(t, again.Pending)
 }
 
-func TestDispatcher_CancelSequenceWithEsc(t *testing.T) {
-	d, err := NewDispatcher([]bindings.Binding{{Action: "git_push", Scope: "revisions", Seq: []string{"g", "p"}}})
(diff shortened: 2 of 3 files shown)


idursun:
Thanks,

but in the linked issue, I thought I made it clear that this is not the approach I would take. This would make `ctrl+[` behave like `esc` globally, not in the context of cancel. Effectively breaking it for users who mapped `ctrl+[` to do something else.

I won't be accepting this PR but it is a good reference implementation for folks who want to have this in their patched version of `jjui`.

PR #735, opened 2026-08-27
feat(files): render file paths relative to startup directory

### LINKED ISSUE #719: FR: Use CWD-relative paths
When checking difference for each individual file, jjui shows the project-relative path for each file. When the path for files are deep and long (e.g., when working on a monorepo), it can be impossible to distinguish which file is selected (especially when working in splitted panes in terminal). 

Use the CWD-relative path instead which is much shorter and thus easier for user to distinguish which file is selected. If concerned about making this behavior default, at least give user an option to configure it. 

Allow user to configure whether to use project-relative or CWD-relative paths.

Describe in the first question.

idursun:
Renders files relative to the current working directory as described as requested in the referenced issue.

image

fixes #719

### DIFF
--- a/internal/ui/fuzzy_files/fuzzy_files.go
+++ b/internal/ui/fuzzy_files/fuzzy_files.go
@@ -87,7 +87,7 @@ func (fzf *fuzzyFiles) updateRevSet() tea.Cmd {
 	path := fzf.selectedPath()
 	revset := fzf.revset
 	if !path.IsEmpty() {
-		revset = fmt.Sprintf("files('%s')", path.Path())
+		revset = fmt.Sprintf("files(%s)", path.Escaped())
 	}
 	return common.UpdateRevSet(revset)
 }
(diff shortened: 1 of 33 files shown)


internal/ui/fuzzy_files/fuzzy_files.go
```diff
@@ -105,9 +107,9 @@ func (fzf *fuzzyFiles) handleIntent(intent intents.Intent) tea.Cmd {
 	case intents.FileSearchCancel:
 		return common.UpdateRevSet(fzf.revset)
 	case intents.FileSearchEdit:
-		path := fuzzy_search.SelectedMatch(fzf)
+		path := fzf.selectedPath()
 		return newCmd(common.ExecMsg{
-			Line: config.GetDefaultEditor() + " " + path,
+			Line: config.GetDefaultEditor() + " '" + path.Path() + "'",
```
baggiiiie: i think this breaks if there's a quote in `path.Path()`

baggiiiie: im seeing in other places there's already something like `file.Escaped()`. maybe we could leverage that here ?

baggiiiie (dismissed):
oh wow the change is so much bigger than i would expect. another question i have is, would this break fuzzy search ? e.g., searching for `internal` may no longer match `internal/ui/ui.go` when it's shortened?

internal/ui/fuzzy_files/fuzzy_files.go
idursun: Thanks. I pushed a commit to address this.

idursun: Yes! and also introduced a `ShellEscaped` version for command execution.

PR #736, opened 2026-09-02
Feat: Add config option to show just the basename in title

### LINKED ISSUE #68: Feature: Show the filepath (or name of cwd) of the current jj repo at the top

nickchomey:
By default jjui shows `jjui - full/path/to/repo` in the title. You can turn that off by setting `ui.set_window_title`  to `false`, which then just shows `jjui`.

This PR adds the ability to set `ui.set_window_title` to `"base"` or `"full"` (which is the same as `true`). 

This results in a significantly cleaner title, especially when many tabs are open and you therefore cant even see the base dir in the title

Before:
Screenshot_2026-09-02_11-23-14

After:
Screenshot_2026-09-02_11-22-49

Seems like it resolves the now-closed https://github.com/idursun/jjui/issues/68

### DIFF
--- a/internal/config/config.go
+++ b/internal/config/config.go
@@ -186,7 +186,7 @@ type UIConfig struct {
 	Theme           ThemeConfig           `toml:"theme"`
 	Colors          map[string]Color      `toml:"colors"`
 	BackgroundBlend BackgroundBlendConfig `toml:"background_blend"`
-	SetWindowTitle  bool                  `toml:"set_window_title"`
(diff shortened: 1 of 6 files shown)


idursun (approved):
Thanks! LGTM.

Issue #733, opened 2026-08-25
Bookmark Pane Feedback Tracking Issue


nickchomey:
@baggiiiie 

Currently we can only multi-select bookmarks in the bookmark pane view by using spacebar. 

It would be helpful to also be able to multi-select bookmarks with alt/ctrl+click, just as we can for revisions in the log. Useful for deleting, tracking, untracking, pushing etc....

baggiiiie:
hey @nickchomey thanks for the issue, feel free to use this issue or open new ones for feedbacks on bookmark pane feature. i'll have a look over the weekend

much appreciated!!

nickchomey:
Ive changed the title to make this a general tracking issue for the bookmark pane. we can split into separate ones if it becomes unwieldy

shyun3:
Seems like panes that overlap the bookmarks pane get cut off:
<img width="848" height="487" alt="Image" src="https://github.com/user-attachments/assets/df0c28e6-d6f9-4bc4-9c86-1c3e26641027" />

To reproduce:
1. Press `b` to open the bookmarks pane
2. Press `tab` to go to the revisions pane
3. Press `g` to open up the git pane

PR #742, opened 2026-09-10
feat: multi-select bookmarks with ctrl and alt click

baggiiiie:
implemented the mouse support mentioned in #733, `ctrl+click` to select , `alt+click` to range select


idursun (approved):
Thanks!

PR #747, opened 2026-09-13
feat: expose preview scroll offset and content to Lua

### LINKED ISSUE #731: FR: Make scroll offset in the preview pane available to Lua functions
Oftentimes when I am reviewing code that I've mostly finished with, I open the detail view with `l` and flip through the files, reading the diff in the preview pane. Occasionally I find some minor thing that I'd like to correct, e.g., a comment that needs rewording. Rather than have to go cross-reference the file in my IDE to achieve this, I've added a Lua function in the detail view so I can press `e` to open the selected file directly in a text editor.

However, if the file is long, it is sometimes annoying to find the place in the file where I wanted to make the change, since the editor shows the entire file rather than just the diff. If it was possible to retrieve an offset or some other kind of scroll information from the detail context showing where I was at viewing the diff of the file, I could grab the line numbers from the diff and pass these to the invocation of my editor, and it could open to the same spot I'm looking at in the preview pane.

Expose the preview pane's current `YOffset` and (ideally, though this could probably be reconstructed manually using `jj diff`) the content of the preview pane in the `context` available to Lua functions running in the detail view. If the preview pane is not open, presumably `nil` should be supplied for both.

I'm not attached to the specific types of information above being available exactly as stated, but I don't think this feature request is resolvable without making some relevant information available in the Lua API.

(It's also not a big deal, just thought I'd suggest this as potential polish.)

As far as I (and Claude) can tell, there is no available way to retrieve this part of the state programmatically.

## Additional context
My current edit function:

```
[[actions]]
name = "edit-file"
lua = '''
local files = context.checked_files()
if #files == 0 then
  local f = context.file()
  if f then files = { f } end
end
if #files == 0 then
  flash({ text = "edit-file: no file selected", error = true })
  return
end

-- Detail-view paths are relative to the workspace root, not to the directory
-- jjui was started in, so anchor them explicitly.
local root = (jj({"workspace", "root"}) or ""):gsub("%s+$", "")
local function resolve(p)
  if root == "" or p:sub(1, 1) == "/" then return p end
  return root .. "/" .. p
end

local function shquote(s)
  return "'" .. s:gsub("'", "'\\''") .. "'"
end

local args = {}
for _, f in ipairs(files) do
  table.insert(args, shquote(resolve(f)))
end

-- Is the change being inspected the working copy? context.change_id() is only
-- the shortest unique prefix (often a single character), so ask jj rather than
-- comparing ids ourselves, and take a displayable id from the same call.
local short, kind
local cid = context.change_id()
if cid then
  local out = jj({"log", "--no-graph", "-r", cid, "-T",
                  'change_id.short() ++ " " ++ if(current_working_copy, "wc", "other")'})
  short, kind = (out or ""):match("^(%S+)%s+(%S+)")
end

exec_shell((os.getenv("EDITOR") or "vi") .. " " .. table.concat(args, " "))

if kind == "other" then
  flash({ text = "Edited working-copy version, not " .. short, error = true })
end
jjui.revisions.details.refresh()
'''
```

mvanhorn:
Add `context.preview_y_offset()` returning an optional integer and `context.preview_content()` returning an optional string, available through the existing `jjui.context` alias too. A user reviewing a file diff in the details view can open the selected file with a Lua editor action, but cannot discover the preview's current position to choose an appropriate editor line.

Preview model: verify the new accessor returns the complete stored content after CR removal and tab expansion, retaining ANSI escapes, blank lines, and trailing newline behavior already used by `SetContent`; vertical and horizontal scrolling do not crop the returned content; Lua API: use `internal/scripting/lua_test.go` to assert exact integer/string values through both `context` and `jjui.context`, explicit nil when the reader is absent or reports hidden, and valid zero/empty values for an active empty preview.

Closes #731


idursun:
Thanks for working on this. 

However, the linked issue is making a suggestion, not specifying an agreed implementation. I’m still discussing the options and haven’t agreed that exposing preview offset/content is the right approach.

mvanhorn:
Understood, and sorry for getting ahead of the discussion on #731, @idursun. I will convert this to a draft so it stays out of your review queue. If the discussion lands on exposing the offset and content, it is ready to pick back up; if you settle on a different approach I am happy to reshape it or close it.

PR #748, opened 2026-09-14
fix(bookmarks): move bookmarks to hidden revisions

### LINKED ISSUE #656: FR: Cannot move a bookmark to a selected hidden revision
I’d like to be able to restore a change or bookmark to an earlier point in its evolog. For example, I have feature@origin and feature*, where feature* was accidentally changed after it had already been pushed. I want to restore it back to the pushed state.
I set the revset to feature@origin and also tried exactly(feature@origin, 1). I can see the @origin bookmark, but when I use b, then m, then filter for feature, jjui uses the change ID instead. Since that change ID is shared across the evolog up to the current head, it resolves to the wrong revision. For this kind of case, I currently have to use a custom command instead.

when moving a bookmark to a selected revision, use the selected commit ID rather than the change ID. Using the change ID is ambiguous across evolog entries and hidden revisions such as <bookmark>@origin.

A clear and concise description of why current features cannot solve the problem.

anandghegde:
Moving a bookmark onto a hidden revision (e.g. `feature@origin` after `feature` was rewritten locally) does nothing: jjui runs `jj bookmark move exact:"feature" --to <change id> --allow-backwards`, the change id resolves to the visible revision where the bookmark already is, and jj replies "No bookmarks to update."

## Cause

`jj.Commit.GetChangeId()` already returns the commit id for hidden revisions, but `Hidden` was never set. The parser looks for a segment that is exactly `hidden`, while current jj (checked 0.37 and 0.45) renders the label as `(hidden)`, so the check never matched.

Once that is fixed, the bookmarks menu (`b`, `m`) showed no move items for a hidden revision, because it lists candidates with `::X | X::` and a hidden revision has no visible descendants.

## Fix

- `internal/parser/row.go`: accept `(hidden)` as well as `hidden`.
- `internal/ui/bookmarks/bookmarks.go`: list movable bookmarks using the change id (same as before for visible revisions), while the move itself still targets `GetChangeId()`, which is now the commit id for hidden revisions.

## Steps to reproduce

1. Push a bookmark `feature`, then rewrite that change locally so `feature@origin` is hidden.
2. `jjui -r 'feature@origin'`, press `b`, `m`, `enter`.
3. Before: `jj bookmark move exact:"feature" --to <change id> --allow-backwards` -> "No bookmarks to update."
   After: `jj bookmark move exact:"feature" --to <commit id> --allow-backwards` -> "Moved 1 bookmarks to ...", and `jj bookmark list --all` shows `feature` and `@origin` on the same commit.

## Testing

- Ran the steps above against a scratch repo with jj 0.45.1, with builds from `main` and from this branch.
- Added `test/testdata/hidden.log` (real jj 0.45.1 output) with a parser test, and a bookmarks menu test for a hidden revision. Both fail without the change.
- `go test ./...`, `go vet ./...`, `gofmt -l` clean.

Fixes #656


idursun (approved):
Thanks!

PR #749, opened 2026-09-14
feat(describe): add clear action (ctrl+x)

baptiste0928:
Added a clear action (defaults to `ctrl+x`) in the describe editor, which kills the whole description to the yank buffer (so it can be restored with `ctrl+y`).

This follows the patterns for the recently added yank support in https://github.com/idursun/jjui/commit/e32d3d. I picked `ctrl+x` instead of the more conventional `ctrl+c` since it is already used as "cancel" in the app. The binding is displayed in the quick help in the bottom.

I decided to directly open a PR instead of making an issue since I already did that change for myself, and I think it can be a good upstream addition. This change has been made with assistance of Claude Fable 5.1 since I'm not familiar with Go, and I reviewed the changes myself before submitting.

Here's a preview video if that's helpful (I'm doing `ctrl+x` then `ctrl+y` in it, and you can notice the added help text in the bottom):

https://github.com/user-attachments/assets/9e1c9178-604d-42a5-bebc-d549772e4b7a


idursun (approved):
Hey, thanks!

I don't know much about the emacs style keybindings, so I searched a little bit but couldn't find default binding for doing the "clear" you implemented. `ctrl+x` collides with the _cut_ shortcut on windows but that's not a problem because we don't have selection support in the input.

I am happy to merge this.

baptiste0928:
Thank you!

